PR #1, opened 2017-01-15
Update Readme.md

blackforestboi:
better installation guide...


Treora:
From our chat: this seemed to be a problem with your `make` configuration on OSX. If this appears a more common problem we could consider ditching `make`&`Makefile` altogether.

PR #34, opened 2017-02-27
[RFC] Introduce CSS modules & covert /src/options over to using them

aquibm:
### Why?
This approach is much less restrictive than using the css-in-js approach that we had earlier. We're simply writing css here and can therefore easily style pseudo elements and make use of css animations. It also brings in new css features (variables, nesting, etc) and follows the golden rule that all styles are local to the component unless explicitly made global. (see routeTitle in base.css)

### Changes:
- The cssOutput prop in the bundle determines whether we'll attempt to build css-modules.
- Stop tracking extension/options/style.css as this is now automatically generated via css-modulesify.
- Covert over all the css-in-js styles in options to use css-modules.
- Add autoprefixing and other features via [cssnext](http://cssnext.io/features/)

### Future Tasks:
- Convert the overview app to use css-modules

### RFC?
This is the first time I've used css-modules through browserify, not sure if it's setup correctly. @Treora, please let me know if I need to change anything 👍


Treora:
I am starting to like the idea of `css-modules`, to organise style in the same hierarchy and to make it traceable where a component's style is defined. It looks like `css-next` gives the main features I wanted to use `scss` for, but now with at least the pretension that it will be normally supported css some day.

If you like this approach I am happy to apply it to the whole project. In any case this seems better than either inline styles defined in javascript or having a single css file.

src/options/components/navigation/index.jsx
```diff
@@ -1,13 +1,22 @@
 import React, { PropTypes } from 'react'
 import { Link } from 'react-router'
-import styles from './styles'
+import classNames from 'classnames/bind'
+
+import styles from './styles.css'
 
 const Navigation = ({ currentLocation, routes }) => {
     function buildRoutes() {
+        let cx = classNames.bind(styles)
```
Treora: From the `css-modules` [docs](https://www.npmjs.com/package/classnames#alternate-bind-version-for-css-modules-) about `classNames.bind`:
> Note that in ES2015 environments, it may be better to use the "dynamic class names" approach documented above.

I agree with them, to avoid this extra level of magic.

src/options/components/navigation/index.jsx
```diff
@@ -1,13 +1,22 @@
...
+
         return routes.map((route, idx) => {
+            let navClasses = cx({
+                navLink: true,
+                isActive: isActive(route)
+            })
```
Treora: so instead of `cx` this would be:
```
classNames({
    [styles.navLink]: true,
    [styles.isActive]: isActive(route),
})
```

src/options/containers/settings/index.jsx
```diff
@@ -1,12 +1,11 @@
 import React from 'react'
 import { Link } from 'react-router'
-import styles from './styles'
 
 class SettingsContainer extends React.Component {
     render() {
         return (
             <div>
-                <h1 style={styles.title}>Settings</h1>
+                <h1 className="routeTitle">Settings</h1>
```
Treora: Instead of making `routeTitle` a global class, would it not be cleaner to `import { routeTitle } from '../../base.css'`?

gulpfile.babel.js
```diff
@@ -47,6 +49,16 @@ function createBundle({entries, output, destination},
     b.transform(envify({
         NODE_ENV: production ? 'production' : 'development'
     }), {global: true})
+
+    if(cssOutput) {
+        b.plugin('css-modulesify', {
+            output: path.join(destination, cssOutput),
+            postcssBefore: [
+                require('postcss-cssnext')
```
Treora: I would import this right at the top rather than require it here. Also 'css-modulesify' could be imported, rather than passed as a string, for clarity and consistency.
Besides that this looks like a neat approach. Runs fine.

src/options/containers/settings/index.jsx
aquibm: That's much nicer, implemented. 👍

gulpfile.babel.js
aquibm: Fixed 👍

package.json
```diff
@@ -35,15 +35,20 @@
     "tldjs": "^1.7.0"
   },
   "devDependencies": {
+    "autoprefixer": "^6.7.5",
```
Treora, line 38: unused?

package.json
```diff
@@ -35,15 +35,20 @@
...
     "babelify": "^7.3.0",
     "browserify": "^13.1.1",
+    "css-modulesify": "^0.27.1",
     "gulp": "^3.9.1",
     "gulp-identity": "^0.1.0",
     "gulp-uglify": "^2.0.1",
     "loose-envify": "^1.3.0",
+    "postcss": "^5.2.15",
+    "postcss-cssnext": "^2.9.0",
+    "postcss-modules-values": "^1.2.2",
```
Treora: Only `postcss-cssnext` is used directly, can the other two be removed without breaking it?

gulpfile.babel.js
```diff
@@ -8,6 +8,9 @@ import watchify from 'watchify'
 import babelify from 'babelify'
 import envify from 'loose-envify/custom'
 import uglifyify from 'uglifyify'
+import path from 'path'
+import modulesify from 'css-modulesify'
+import cssnext from 'postcss-cssnext'
```
Treora, line 13: nitpicking: I'd keep the whole name and camelCase it as `cssModulesify`, especially since `modulesify` gives no hint what it does. Not sure between `cssnext`/`postcssCssnext`.

Treora:
Apart from these details, would this be ready to merge?

Treora:
By the way, feel free to squash all the tiny fixes in a single commit, keeping up the appearance that everything was perfect at the first shot. ;)

package.json
aquibm: This seems to be an weird peer dependency of cssnext. Build fails without it.

aquibm:
Yep, this is good to merge :shipit:

PR #35, opened 2017-02-28
Introduce css-modules to the overview app

aquibm:


aquibm:
Hey, @Treora, heads up if you have any css changes locally:
Once merged, this will stop tracking `extension/overview/style.css` as it'll now be built at compile time.

src/overview/components/Overview.jsx
```diff
@@ -5,11 +5,13 @@ import * as actions from '../actions'
 import { ourState } from '../selectors'
 import ResultList from './ResultList'
 
+import styles from './Overview.css'
```
aquibm, line 8: I've created the styles in the root components directory for now since we only have three components, I'd imagine that if we wish to add more each component could be nested in its own folder with its own styles.

src/page-viewer/styles.css
```diff
@@ -0,0 +1,6 @@
+.root {
+    float: right;
+    padding: 4px 4px 4px 12px;
```
Treora: I feel the float and padding styles should not be defined in this component; they specify how to place the component in its context, not how the component itself looks.

Not sure what is a good practice for splitting this up, I guess CSS was not designed with modularity in mind. Maybe we should keep this component purely technical, i.e. not do any styling in here. The overview can just pass `className` down as a prop.

Treora (commented):
Nice, thanks.
The styling itself could probably use some attention too, but that's for another time..

src/page-viewer/styles.css
aquibm: That sounds like the best way to do it, I think common practice is to setup your components with some reasonable defaults and have everything else passed in by the caller.

But yeah I agree that we pass through float and padding in this case as it's to do with the position of this component relative to its parent.

aquibm: Looks like we'll either have to pass all the styles down, as passing down the className prop will overwrite the component's own className or do something like this:

```jsx
export const LinkToLocalVersion = ({page, children, ...props}) => {
    let classes = classNames({
        [styles.root]: true,
        [props.className]: true
    })

    return (
        <a
            href={`data:text/html;charset=UTF-8,${pageToHTML(page)}`}
            title="Stored text version available"
            {...props}
            className={classes}
        >
            {children}
        </a>
    )
}
```

What do you reckon @Treora?

Treora: Merging the `props.className` with its own `className` looks like a logical solution. I think however that in this particular case the component does not really need reasonable defaults to be set, as a normal unstyled link with underline is a reasonable default too (of course an underline would look weird on the unicode symbol we use, but that symbol is passed as `children`, so its style is not up to the component).
In short: just put the whole style in overview.

aquibm: Sounds good 👍

Issue #27, opened 2017-02-14
Extract text content from PDFs


Treora:
Mainly just making a cleanup of Oliver's code: https://github.com/oliversauter/pdf2text4extensions

bohrium272:
I'd like to take this up. Please describe further.

Treora:
Great. I imagine this to be a separable module, though possibly developed within this repo for now, e.g. in `/src/pdf-analysis`.

As we quickly discussed, there may be two situations to consider: one is where the user is visiting the document, one is where we just have the url and should download it first. I am not sure about the possibilities in the first case, we might be required to download it again if we cannot access it otherwise. Anyway, we would probably end up with a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) containing the document.

So let's start with making a single function that, given this pdf blob, returns an object with its plain text and metadata, for example like so:
```
>>> analysePdf(myPdfBlob)
{
  plainText: "bla bla blabla...",
  title: "Bla.",
  numPages: 3,
  ...
}
```

We'll probably use the core of [PDF.js](https://mozilla.github.io/pdf.js/) in the same way [Oliver's code](https://github.com/oliversauter/pdf2text4extensions) does, though I am not sure if we can import just the core or if it only comes along with the rendering part. Perhaps have a look at its docs and examples.

oliversauter:
I talked to @vaibhav, he wanted to take this task on. Is that still valid Vaibhav?

@Treora I suggest adding him as a contributor, so we can assign the task.

blackforestboi:
I talked to @vaibsharma, he wanted to take this task on. Is that still valid Vaibhav?

@Treora I suggest adding him as a contributor, so we can assign the task.

vaibsharma:
Yes I want to take this task .

blackforestboi:
ok great. Happy Hacking!

Treora:
@vaibsharma / @oliversauter: do I understand plans changed again? For me this feature does not have a high priority and nothing is blocked by it, so no problem if it moves to the waiting list again, but if somebody wants to implement it I am glad to assist.

blackforestboi:
@vaibsharma is having his midterm exams until 11 March, so wont be able to take it on anytime soon and there are other people who want to take it on...soon coming :)

vaibsharma:
I pushed whatever I was able to commit . Please have a look at #43 . Rest I will do after midsems :) . Thanks

[...]

PR #43, opened 2017-03-02
Pdfgen

vaibsharma:
This is how I thought we should extract texts from the pdf , and all the scripts for pdf extraction are inside the pdf-analysis . Is that how you wanted to the files for pdf extraction should be @Treora . I am working on the extraction in pdfgen file, will commit asap .


Treora:
Hi Vaib, thanks for sharing your progress. I understand this is a snapshot of an unfinished effort.

> Is that how you wanted to the files for pdf extraction should be @Treora

You are right to call performPdfAnalysis instead of performPageAnalysis, but I did not intend to create a near full copy of page-analysis in the pdf-analysis. As [said](https://github.com/WebMemex/webmemex-extension/issues/27#issuecomment-280873617), we could start with just a function that grabs the plain text from a pdf blob, and then add a function to first fetch the file from the web.
Also, the pdfjs library should not have to be inside our source tree; perhaps the [build on npm](https://www.npmjs.com/package/pdfjs-dist) is suitable to import.

vaibsharma:
I have now build a pdfjs-dist package of npm for now and yes I understand the sameness of the code , it was because I thought you wanted an entire different files for checking them , but we can do that in either way no problems .

Treora:
Closing this; we'd better tackle this later and take a cleaner approach.

PR #47, opened 2017-03-02
Added : import-bookmarks.js

### LINKED ISSUE #29: Import browser bookmarks
Get all of the browser bookmarks and add them to the database, while marking them as being imported.

Each bookmark would map to a page in the database. Perhaps we should also try to pull the visits to these pages from the history, even if one does not choose to import the full history.

How to represent that a page is bookmarked is a remaining question. Adding a flag to the page itself sounds easy but is perhaps the wrong place. Adding a new type of item that links to the page is probably a more logical solution, to keep the pages themselves separated from the user's organisation.

The folder structure, tags and descriptions of bookmarks are important, but perhaps we will deal with those only later on.

wimpykid26:
Will import all browser bookmarks
and store them in db.

Fixes #29


wimpykid26:
@Treora @oliversauter wrote a new file which is somewhat similar to import-history.js. I am having a great deal of trouble in testing/debugging it. Could you please guide me on how do it. The console is cluttered up when the overview.html opens therefore its of no use to create a button there and call it. Thanks in advance :)

Treora:
Great you started on this.

By the way, if a lot of code is the same as in `import-history.js`, we could consider factoring those parts out or `export`ing them from either module so they can be shared.

As for debugging, add `import './browser-bookmarks/import-bookmarks'` to `src/background.js`, so that the file is included in the background. You should get a completely blank console if you click 'inspect views: background page' in the list of extensions. There you could test your functions if you add them to `window` as described before, e.g. `window.importBookmarks = importBookmarks`.

wimpykid26:
@Treora . Made some changes and debugged the code. Its working perfectly fine with custom inputs as bookmarks. `browser.bookmarks.getTree()` and `browser.bookmarks.getRecent()` return null object arrays when called. Even after bookmarking various sites in the browser that opens while running the extension. Cannot seem to get my head around that. Please review. Thanks :)

blackforestboi:
Hey @wimpykid26 

Don't take my word for it, but when I implemented 'browser.bookmarks.getRecent()' I had to add a few parameters. Here you see the old code, where I imported the bookmarks: https://github.com/WorldBrain/Research-Engine/blob/master/src/js/import_bookmarks.js#L11

Looking into [the API documentation](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/bookmarks/getRecent) it says: 
> var gettingRecent = browser.bookmarks.getRecent(
  numberOfItems          // integer
)

> numberOfItems
A number representing the maximum number of items to return. The returned list will contain up to this many of the most recently added items. The minimum allowed value here is 1. If you pass 0 or less, the function will throw an error.

Hope that helps!
Oliver

wimpykid26:
Really appreciate the review @oliversauter.

I did actually pass an integer inside the function call. Still it didn't return anything. There is nothing wrong with the function call. It's just not returning anything. Probably it's not keeping track of the bookmarks in the browser its opening after `fx-run`

wimpykid26:
I'll look into it some more and keep you updated . Thanks :)

src/browser-bookmarks/import-bookmarks.js
```diff
@@ -1,143 +1,148 @@
...
+  }
 
-function getBookmarkItems() {
-  var bookmarkTree = browser.bookmarks.getTree();
-  return bookmarkTree.then(logtree(bookmarkTree[0]), onRejected);
-}
+  function getBookmarkItems() {
+    var bookmarkTree = browser.bookmarks.getRecent();
+    console.log("inside getBookmarkItems");
+    return bookmarkTree.then(logtree(bookmarkTree[0], 0), onRejected);
```
Treora: `bookmarkTree` is a promise, it will give you the value only later. You should pass `then()` a callback function, which will be called with the result of the promise (the bookmarks in this case) as soon as that result is available.
Perhaps read a tutorial on javascript Promises, and have a look at other examples in the code.
Try something like `bookmarkTree.then(result => console.log(result), onRejected)`. In case you are not familiar with ES6, `arg => returnValue` is just short for `function (arg) {return returnValue}`.

wimpykid26:
screenshot from 2017-03-07 19-09-12
Checked if url is valid via `isWorthRemembering`
converted valid bookmarks to allDocs (pageDocs + visitDocs)
stored them in db.

Please review @Treora @oliversauter

Treora:
@wimpykid26: Either I have not understood what you are doing, or there is still a load of unnecessary complexity in there which would take much more time to point out than to rewrite (sorry to be harsh), so it does not seem ready for review. As said before, I think the tree should not be necessary at all, you could just getting the last zillion recent bookmarks with `browser.bookmarks.getRecent` and then do nearly exactly the same things the history import does, probably sharing some code with it too.

wimpykid26:
@Treora Apologies for the complicated code. Removed the `logtree` function (which I had used for `browser.bookmarks.getTree() ` but later forgot to remove) and imported reusable functions from import-history.

src/browser-bookmarks/import-bookmarks.js
```diff
@@ -0,0 +1,41 @@
...
+   convertHistoryToPagesAndVisits } from '../browser-history/import-history'
+import { isWorthRemembering, generatePageDocId, generateVisitDocId,
+  visitKeyPrefix, convertVisitDocId } from '../activity-logger'
+
+    function onRejected(error) {
+      console.log(`An error: ${error}`);
+    }
+
+    function getBookmarkItems() {
+      var bookmarkTree = browser.bookmarks.getRecent(100000);
```
Treora: `bookmarkTree` is not the right name anymore. It does not really need a name though, you can just return the whole promise chain directly.

Treora: Also, though a 100k bookmarks sounds insane, let's go a bit further on the insane side. `importHistory` uses 9999999 as limit, could make it the same here.

src/browser-bookmarks/import-bookmarks.js
```diff
@@ -0,0 +1,41 @@
...
+      return bookmarkTree.then(bookmarkItems =>
+          bookmarkItems.filter(({url}) => isWorthRemembering({url})));
+    }
+
+    export default function importBookmarks() {
+      return getBookmarkItems().then(result =>
+      getVisitItemsForHistoryItems(result)).then(
+          res => {
+            var docs = convertHistoryToPagesAndVisits(res)
```
Treora: Does this actually work without problems? You are passing bookmark items as if they were history items twice here (line 20&22). I suppose that it could work as they also just have a `url` and and `id` field, and that is all we use (I think?). We should probably make these functions from the history import a bit more generic to also work for bookmarks, e.g. `transformPageDoc` should probably set a different nonce depending on which import process uses it. Not sure whether we would want to deduplicate or link up bookmarks and history items when both are imported. Needs thought.

Treora (commented):
Great, this is much more like it.

It is nice that so much code can be reused from history. However, I realise now that this may actually require some more thinking. You are treating bookmark items as if they were history items, and they may indeed be similar enough for this, but we should probably make the reused functions more generic (revised to be less history-specific). Also there are questions like whether a bookmark and a history item should be deduplicated if both are imported. I will have to think about this, and first merge the deduplication feature before continuing on this one.

src/browser-bookmarks/import-bookmarks.js
wimpykid26: @Treora This does work smoothly as the bookmarks and the history hasve the same object structure. As you said , the only difference I could think of between both of them is setting the nonce according to which import uses it. To make the functions generic, I could move them to a seperate file and rename them and import them in the history and boomarks code. We could either insert a flag value into the model maybe?.

wimpykid26:
@Treora how do I set the nonce differently. getVisitItemsForHistoryItems eventually finds the last visit of a bookmarkItem and returns a historyItem therefore only historyItems are getting passed into generatePageDoc. For retrieving the visitItems the bookmarkItem has to be passed through getVisit...... . Instead should I bypass for bookmarks and get the `dateAdded` property for its last visit time?.

Treora:
I'm sorry I have not yet got around to follow up on this; other priorities.. It would be nice to continue with the last refactoring one of the next weeks, though as said it needs a bit of thought what the desired result actually is. It should become a bit easier to refactor after PR #81, which converts the promise chains to `async`/`await` syntax and in the process also refactors history import code a bit.

poltak:
I've been working on getting up-do-date with all these import-history/bookmarks + fetch/analyse of pages issues for the WorldBrain fork ([the ones all mentioned here](https://github.com/WebMemex/webmemex-extension/issues/29#issuecomment-295509334)). import-bookmarks seems like a good place to start, as it shares a lot with the existing `import-history` code, and some progress has already been made here. 

At the moment, looking over the @wimpykid26's code, the `import-bookmarks` stuff creates visit and page docs same as was done in `import-history`. If my understanding is correct, after reading through #29, the main difference here should be that we'll want to save some kind of bookmark doc instead of a visit doc. Is my understanding correct? 

So:
- `import-history` module: grab everything from history, generate __visit__ + page docs
- `import-bookmarks` module: grab everything from bookmarks, generate __bookmark__ + page docs

If so, let's decide some structure for bookmark docs? 
Visit doc structure currently looks like: 
```
{
  _id, 
  url, // URL of the visited page
  visitStart, // timestamp of when the site was visited
  importedFromBrowserHistory, // timestamp of when the import happened
  referringVisit: { _id }, // optional ID of other visit doc deemed to be referrer for this visit
  page: { _id }, // ID of the page doc which this visit is associated with
}
```

Bookmark doc structure could possibly look like: 
```
{
  _id, 
  url, // URL of the bookmarked page 
  bookmarkStart, // timestamp of when the site was bookmarked
  importedFromBrowserBookmarks, // timestamp of when the import happened
  page: { _id }, // ID of the page doc which this bookmark is associated with
}
```

Do we need any additional information about bookmarks? At the moment, according to discussion in #29, we're leaving out things like folder structure. Only difference now is there's no optional referrer visit reference, because it's not applicable here, and renaming of certain atts.

An additional unrelated question: When will the `url` in a visit doc and the `url` in that visit's associated page doc differ? All the URLs in the associated docs i'm looking at in my local pouchDB seems to be the same.

Treora:
As most of the discussion is not specific to this PR but referring to the issue discussion, I replied [in issue #29](https://github.com/WebMemex/webmemex-extension/issues/29#issuecomment-296225864) to keep things together.

Treora:
Closing due to inactivity. A more likely implementation for this feature is being made by @poltak over at WorldBrain (see https://github.com/WorldBrain/WebMemex/pull/25).

PR #49, opened 2017-03-04
Correct the link for util/webextensionRPC.js

Chaitya62:
Nothing major there was a linking error in util/webextensionRPC.js thought it should be correct!


Treora:
Thanks!

Issue #40, opened 2017-03-01
Overhaul overview (search results)


blackforestboi:
The current search results are somewhat limited in the amount of info to be able to see/skim. The space could be used more efficiently. 

## The issues I see:
1. Space not used, so no place to show additional important info for the search results:
![screen shot 2017-03-01 at 21 57 32](https://cloud.githubusercontent.com/assets/7870039/23480988/31ce56c0-feca-11e6-8ee3-4c2fe4efebbd.png)

2. The title-background is basically covering half of the screenshot, making it only half as useful. (no pun intended) :)
<img width="260" alt="screen shot 2017-03-01 at 21 56 57" src="https://cloud.githubusercontent.com/assets/7870039/23480946/06c23668-feca-11e6-847e-19d1d9757c80.png">

3. Only info shown is a icon for the page (which is not always there) and a very short excerpt of the title + half the screenshot.

4. The groupings group pages from similar domains, I dont see that as very useful, especially in the search results, because I have to do one more step in order to find out which one is mine. If I am browing on the same page a lot (GitHub could be one) I have to hover over each element in order to find out which one was the correct.

5. The elements indicating the pages visited before and after a search result take too much space and are not really clear to which they belongs, or if it is a continuation. (Ultimately its not clear to the user what they are at all)

## What I propose:
A combination of what is currently in place and what @aquibm developed a few weeks back (with screengrab-gif): https://github.com/WorldBrain/Research-Engine/pull/91

So the combination would be to have the screenshot on the left, with title and excerpts of the results page on the right, making the results about 2,5/3 times as wide as they are now. 

I would throw into the room of removing the groupings with the layers behind each other. And just let it be a list of elements in the search results. 

And lastly making the before and after pages simply a list of titles that maybe expand to the whole element by hovering over them. Also adding some indication that its what came before and after. 

@Treora @aquibm @obsidianart 

What do you guys think?

Treora:
The current design is just what I quickly developed to get something usable, but I would be happy if somebody created something nicer. There is a reason for the empty space and small size of the items: the intention is to create a two-dimensional graph layout rather than a single column as it is now. As this may not materialise soon, it would probably be good to make some quick improvements on the current single-column design.

[...]

PR #55, opened 2017-03-06
Issue #40 -- Overhaul overview (search results)

reficul31:
This is in reference to the [issue](https://github.com/WebMemex/webmemex-extension/issues/40).
screenshot from 2017-03-05 20 16 29


Treora:
I just tried it out, this design looks more informative than the current squares. A bunch of small things I would change still (some were already discussed in #40):
- more space around the buttons
- remove clustering, or at least the styles for popping up clustered items on hover
- change to ellipsis overflow for titles & urls (also they do not get to use the full width for some reason?)
- "No screenshot available" is not so important, could be very thin and lightgrey, or removed completely.
- keep the background image until we have a better one

And possibly some other small things I don't notice now.

reficul31:
Thanks for the review. I'll change right now.

reficul31:
Updated PR:
1. Added significant spacing between the buttons.
2. Removed the clustering from the ResultView
3. Changed text-overflow to ellipsis
4. "No Screenshot Available" made thinner and more lighter
5. Background image back up

Please review and tell me what else should be changed. Thanks for your time.

Treora:
Nice. Sorry to keep commenting, but I just played around with it more, finding that the use of relative sizes everywhere has its problems. E.g. a small window size just looks very wrong:

relative-sizes-problem

The layout seems to be designed to be nice on a normal screen size, but that is not the right way to design things. I see you also use `display:table` for vertical alignment, and float left and right , and wondered if you know that nowadays there css has `flex` layout. It takes a moment to get into it though, but allows making such things nicer and more.. flexible. I can also have a try at improving it some moment, or maybe somebody with more design skills than I can.

reficul31:
I am so sorry I forgot the min-height of the component(rookie mistake). I will try and be more careful next time. Obviously that doesn't solve the responsive problem. I have changed the PR. I looked into flex and it seems it solves some of our responsive problem. I will change the code again and make the PR. Thanks a lot for pointing out the mistakes. Any opportunity to learn is a good one. :D

Chaitya62:
Just a suggestion but why don't we add a placeholder icon when we cannot get the fa
Vicon instead of that white page icon.

src/overview/components/ResultList.css
```diff
@@ -2,7 +2,7 @@
     list-style: none;
     margin: 40px auto;
     padding: 0;
-    width: 242px;
+    width: 60%;
```
obsidianart: unless you have a container above in pixel you need min-width too

src/overview/components/VisitAsListItem.css
```diff
@@ -1,11 +1,11 @@
 .root {
     position: relative;
     display: block;
-    width: 240px;
-    height: 120px;
+    width: 100%;
```
obsidianart: any particular reason for this? width is auto by default, which means 100% already for most scenarios.

src/overview/components/VisitAsListItem.css
```diff
@@ -1,11 +1,11 @@
...
+    height: 10%;
```
obsidianart: is this 10% of which element? if you don't know probably you want to use wh instead.

src/overview/components/VisitAsListItem.css
```diff
@@ -38,36 +38,56 @@
...
-    top: 0;
-    width: 100%;
-    padding: 8px 0;
-    background-color: rgba(255, 255, 255, 0.9);
-    border-top: 1px solid lightgrey;
-}
-
-.title {
+.logo {
+    height: 100%;
```
obsidianart: as above, are those percentages of which element?

src/overview/components/VisitAsListItem.css
```diff
@@ -38,36 +38,56 @@
...
+    width:10%;
     float: left;
     max-width: 100%;
     overflow: hidden;
+    color: #333;
+}
+.title{
+    height: 100%;
```
obsidianart: as above, are those percentages of which element?

src/overview/components/VisitAsListItem.css
```diff
@@ -38,36 +38,56 @@
...
+    margin:0;
+    padding:0;
+    float:left;
+    text-decoration: none;
+}
+.title div{
     white-space: nowrap;
+    margin:0;
+    padding:0;
+    height:33.33%;
```
obsidianart: as above, are those percentages of which element?

src/overview/components/VisitAsListItem.css
```diff
@@ -38,36 +38,56 @@
...
+    height:100%;
+    margin-right:0.5%;
+}
+.linkToLocalVersion img{
+    margin:40%;
+}
+.screenshot{
+    float:right;
+    width:15%;
+    height:100%;
```
obsidianart: as above, are those percentages of which element?

obsidianart:
I see a large use of percentages for sizing. Those are cards and they don't really make sense in percentage sizing for height.
Percentage is a very tricky thing to use, the value depends on the first parent specified in pixel for width and height but on the element for padding (and probably margin), and you can't move the card from one place to another without risking to break it. I don't think they percentages are needed to style those cards.
About the 3 columns system, I would probably evaluate flexbox rather than floating, since this is for modern browser, but this can open more issues and it can be done later.

reficul31:
Thank you @obsidianart for the review. Yeah as pointed by @Treora I was going to switch over from percentages and move to more definite sizing scheme. Currently I am in the process of converting all the floats to flex. The PR shall be updated soon. Sorry for the css practises.

reficul31:
Added the initial implementation of the above designed front end with flex box css. I have tried to conform to the styles mentioned in the previous PR that I made. 
@obsidianart @Treora please review the PR and tell me all the changes to be made.

src/overview/components/VisitAsListItem.css
```diff
@@ -33,41 +35,54 @@
 
 .thumbnail {
     display: block;
-    max-width: 240px;
-    width: auto;
-    height: auto;
+    height: 100%;
+    width: 100%;
```
obsidianart: do you need this?

obsidianart:
I still see various size in percentage, is there a reason for it?

src/overview/components/VisitAsListItem.css
reficul31: It was observed when the page width was decreased the div with the screenshot was shrinking more when no screenshot was present and less when there was a picture in the div(screenshot was present). 
I researched a bit more and found that when the width of the container is decreased in flex, it fits according to the content given in the div. Therefore the height and width are set to 100% for when there is no screenshot the div occupies the same space as that of when there is a screenshot.

reficul31:
I have tried my best not to use percentages in sizes. I think now whatever percentage sizing being used is necessary(according to me). 
If any mistakes are found I'll change and make the PR again. Thanks for the review! :D

reficul31:
@Treora @obsidianart It recently came to my notice that this PR was never merged(Sorry). I have made the PR merge ready again. We can either close it and go with Oliver's [approach](https://github.com/WebMemex/webmemex-extension/pull/82) or I can change the PR again. Thanks.

blackforestboi:
@reficul31 the main changes I (wanted to make) were regarding the font-size of the different lines, the spaces between elements and some other cleaning up of CSS. 
See the screenshot. 

screen shot 2017-04-02 at 12 08 16

Treora:
I think Oliver's approach is just this PR plus some changes to the design, but he did not continue from this branch so @reficul31's commits disappeared from the history.

I'd take this branch as it was, rebase onto master (rebasing is cleaner than merging), then apply Olivers changes to it (again a rebase), and end up with a single PR with all your edits combined and attributed to the right authors.

reficul31:
I have rebased the master on this branch as per @Treora advice. I have implemented the design by @oliversauter on this branch as well. Please tell me any further changes are required.
@oliversauter thanks for the design input. It looks much cleaner now. Although I have left the background as is due to the earlier conversations on this thread. Here is how the final design looks like.
screenshot from 2017-04-05 00 11 28
edit: removed the merge conflicts. PR is merge ready. :)
Please tell me if any more changes are required in the code or the design.

Treora:
Rebased and merged (de16361).

@reficul31: One thing to I would like to check still: are the icons you used in the public domain, not copyrighted? Also, are you okay with waiving your copyrights on your contributions? I try to make&keep this repo completely copyright-free.

reficul31:
>  Also, are you okay with waiving your copyrights on your contributions? I try to make&keep this repo completely copyright-free.

Yes, I willingly waive my copyrights on my contributions.

>One thing to I would like to check still: are the icons you used in the public domain, not copyrighted?

Yes, they were as far as I remember. The empty screenshot icon was give to me by @oliversauter. I'll do a check on it once.

Issue #54, opened 2017-03-06
Create a 'loading results' indicator.


reficul31:
While the results are being loaded I think we could add an asynchronous call that issues a "loading results" action. It can basically display that results are being loaded or we can even add a progress bar.

Treora:
I think the required actions are already there ([refreshSearch](https://github.com/WebMemex/webmemex-extension/blob/1423c9333112bf8f870764483bca2de14e250e9a/src/overview/actions.js#L28) and [setSearchResult](https://github.com/WebMemex/webmemex-extension/blob/1423c9333112bf8f870764483bca2de14e250e9a/src/overview/actions.js#L11)), but (1) they should toggle a flag in the state to indicate that search is busy, and (2) the overview should read this flag and show a nice loading indicator when appropriate. I can fix the first task if somebody wants to pick up the second (or feel free to do both).

reficul31:
@Treora 
> I think the required actions are already there (refreshSearch and setSearchResult), but (1) they should toggle a flag in the state to indicate that search is busy, and (2) the overview should read this flag and show a nice loading indicator when appropriate.

Or we could make a new epic that triggers a new reducer that changes the state to loading. And when the results are ready we change the state back to display. 

I would like to take up both the tasks. I think the approach to follow is
1. Start by making animation for loading.
2. Set state and trigger animation.
3. Add an epic that triggers the state to loading.
4. Once the searchResult var is ready change the state to ready.
5. Stop animation and display results

I hope I didn't misinterpret.

obsidianart:
I would change 1 in "Create a component for loading", it's very easy to drop it with something like
```
if (search.isFetching) return <Loading />
```

reficul31:
@obsidianart 
Just for clarification do you want loading to be a dumb component?
In my opinion I think making something like this would be cleaner(?)

```{search.isFetching?<div class='loading></div>:null}```

I'm sorry if I misunderstood you completely.

Treora:
Epics work too, but may make it more complicated. I would say try write
it out and see what results in the most elegant and understandable code.

As for starting/stopping an animation; I assume it could just be a
simple `{searchIsLoading ? <some-loader-icon-component /> : null}`. Not
sure where, next to the query input perhaps? We could also colour the
input's background slightly differently as an indicator. A progress bar
is impossible as we do not know the progress.

One thing to note by the way: it can sometimes happen that multiple
searches are running in parallel if one took a long time and the query
[...]

PR #56, opened 2017-03-06
Add no result found message issue#53

Chaitya62:
I added a simple if statement and a message style in ./ResultList.jsx and ./ResultList.css

### DIFF
--- a/src/overview/components/Overview.jsx
+++ b/src/overview/components/Overview.jsx
@@ -21,7 +21,7 @@ class Overview extends React.Component {
             </input>
             {this.props.waitingForResults
                 ? <LoadingIndicator />
-                : <ResultList searchResult={this.props.searchResult} />
+                : <ResultList searchResult={this.props.searchResult} searchQuery={this.props.query} />
             }
         </div>
     }
--- a/src/overview/components/ResultList.jsx
+++ b/src/overview/components/ResultList.jsx
@@ -26,7 +26,17 @@ const timeGapToSpaceGap = makeNonlinearTransform({
     nonlinearity: Math.log,
 })
 
-const ResultList = ({searchResult}) => {
+const ResultList = ({searchResult,searchQuery}) => {
+    //check if there are no results returned
+    const noResultMessage = 'No result found';
+    if(searchResult.rows.length === 0 && searchQuery !== ''){
+        return(
+            <p className={styles.noResultMessage}>
+                {noResultMessage}
+            </p>
+        );
+    }
+
     // The space and possibly a time stamp before each row
     const rowGaps = searchResult.rows.map((row, rowIndex) => {
         // Space between two rows depends on the time between them.
--- a/src/overview/components/ResultList.css
+++ b/src/overview/components/ResultList.css
@@ -28,3 +28,22 @@
         transform: translate(-5px, -75px) rotateZ(-5deg);
     }
 }
+
+.noResultMessage{
+    text-align: center;
+    font-size: 30px;
+    animation-name: fadeIn;
+    animation-duration: 0.5s;
+    animation-iteration-count: 1;
+    animation-timing-function: ease-in;
+}
+
+@keyframes fadeIn {
+    0% {
+        opacity: 0;
+    }
+    100% {
+        opacity: 1;
+    }
+}
+


src/overview/components/ResultList.jsx
```diff
@@ -27,6 +27,15 @@ const timeGapToSpaceGap = makeNonlinearTransform({
 })
 
 const ResultList = ({searchResult}) => {
+
+    //if no result found display relavant message
+    if(searchResult.rows.length === 0){
+        return <p className={styles.message}>
```
obsidianart: unless it is a project decision what do you think of the style (I've seen it most often because it keeps opening close lines)
```
return (
    <p className={styles.message}>
        No result found
    </p>
)

Comment (Chaitya62) [reply src/overview/components/ResultList.jsx]: Sorry didn't get you? is it about the way I wrote the code or the css styling

Comment (obsidianart) [reply src/overview/components/ResultList.jsx]: only about indentation, and only if you think my way is better, and only if my way doesn't violate the rest of the project. To show another big project on this style https://github.com/acdlite/redux-router/blob/master/examples/basic/index.js#L49 (line 49)
I normally don't put weight on style but I started like that with the return and it made my life much harder when the project grew

Comment (Chaitya62) [reply src/overview/components/ResultList.jsx]: @obsidianart okay got it! but the ul just below that had 
```javascript

   return <ul className={styles.root }>
     .
     .
    </ul>

```

So I also did the same thing but yes we can surely choose one of the way to write it and include it somewhere in the project description Readme.md or coding guidlines page in the future.

Chaitya62: I ll wait for @Treora 's input in this will change it if necessary ! though I completely agree  with you @obsidianart  and yes I am also picky about coding styles at times it's doesn't matter at the early stage but helps a lot later on.

Treora: This is one of those stylistic choices where I sway back and forth myself. I'd say keep it either way you like now, and we will restyle the whole codebase at some moment if we think it's better. Perhaps combined with a switch to indenting with only two spaces, I just use four because of old Python habits, but it may be less appreciated in the general javascript etiquette.

src/overview/components/ResultList.jsx
```diff
@@ -27,6 +27,17 @@ const timeGapToSpaceGap = makeNonlinearTransform({
 })
 
 const ResultList = ({searchResult}) => {
+
+    //if no result found display relavant message
+    if(searchResult.rows.length === 0){
+        return (
+            <p className={styles.message}>
+            no result found
```
Treora: Talking about style, please do indent this line too. Also "no results found" sounds slightly wrong, I would either say "No results", or "No pages found" or something like that.

I now realise it would be good to make a special case for when the database is empty. "No results" is not a nice welcome message on a fresh install. Perhaps move the current code to the Overview instead of the ResultList, then test for both the length of searchResult and the existence of a query, and if both are empty show something nice, e.g. "This is your digital memory, watch it grow as you browse the web!"

Chaitya62: Cool I ll make the necessary changes

src/overview/components/ResultList.jsx
obsidianart: I'm very free about style. I don't care about semicolon, spaces, and many other things. This is one of the few things I had a real issue with but the final choice for me is your.

Chaitya62:
In the 4th commit I added a new method in Overview class though I am not proud of what I named it are there anyother suggestions? I could not come up with a better option.

src/overview/components/Overview.jsx
```diff
@@ -18,15 +34,18 @@ class Overview extends React.Component {
                 ref='inputQuery'
             >
             </input>
-            <ResultList searchResult={this.props.searchResult} />
+            {this.createResultList()}
```
obsidianart: this should probably be a component not a function.
I had the same approach at the beginning but the idea of JSX is towards composition.
I don't know if the current code structure allows 2 elements in the same page.
otherwise, the quickest
```
{isFirstVisit && <p className={styles.message}>This is your digital memory, watch it grow as you browse the web! </p>}
{!isFirstVisit && <ResultList searchResult={this.props.searchResult} />}
```
Again, that's my opinion, and React has many opinions, so the choice is yours

Chaitya62:
I didn't want to make a new component just for a special case maybe If not function the second approach you pointed out is a feasible solution

Treora:
> |{isFirstVisit && <p className={styles.message}>This is your digital
> memory, watch it grow as you browse the web! </p>} {!isFirstVisit &&
> <ResultList searchResult={this.props.searchResult} />} |
>

Two tests? Probably it is not the cleanest solution either, but this
became my usual style in this project:

```
{
    isFirstVisit
        ? <p>...</p>
        : <ResultList .../>
}
```

But if the lines get too long, perhaps define the message as a string
constant up in the file.

Interesting how many styles there are indeed.. and how it seems one of
those things where humans are more fussy about than computers. :)

Chaitya62:
Made the changes !

Treora:
Just curious where the original message went now, the one telling there are no results.. ;)

src/overview/components/Overview.jsx
```diff
@@ -8,7 +8,16 @@ import ResultList from './ResultList'
...
+
+
     render() {
+
+        //check if the user visited for the first time
+        const isFirstVisit = (this.props.query === "" && 
+                              this.props.searchResult.rows.length === 0);
+
+        //message for first visit 
+        const message = 'This is your digital memory, watch it grow as you browse the web!';
```
Treora: Too generic name. `firstVisitMessage` perhaps?

src/overview/components/Overview.jsx
```diff
@@ -18,7 +27,16 @@ class Overview extends React.Component {
...
+            {
+                isFirstVisit
+                ? 
+                <p className={styles.message}>
+                    {message}
+                </p>
+                    
+                : <ResultList searchResult={this.props.searchResult} />
+            }
+
```
Treora: I don't know an actually nice style for these things, but for consistency you could keep the indentation that the rest of the code used:
```
isFirstVisit
    ? <p>..</p>
    : <ResultList ...>
```

And if the `<p>` component should take multiple lines:
```
isFirstVisit
    ? (
        <p ...>
            {...}
        </p>
    )
    : <ResultList ...>
```

src/overview/components/VisitAsListItem.css
```diff
@@ -70,4 +70,4 @@
     padding: 4px 4px 4px 12px;
     color: inherit;
     text-decoration: none;
-}
+}
```
Treora: You can leave that newline in its place, I think it was living quite happily there.

src/overview/components/ResultList.jsx
```diff
@@ -27,6 +27,18 @@ const timeGapToSpaceGap = makeNonlinearTransform({
 })
 
 const ResultList = ({searchResult}) => {
+
+    //if no results are found
+    const message = 'no result found';
```
Treora: Too generic name

Treora (commented):
Looks good; just some small style stuff:
- keep and add newlines at end of files
- give more meaningful names to the constants
- some indentation

Treora:
Ok, looks ready to merge? I'll nitpick on some last whitespace myself.

One question still: are you okay with waiving all your copyright on this code, so that the whole code base can be published in public domain? (see unlicense.org)

Chaitya62:
Yes no problem at all.

Treora:
Hmm I just tried it out, and it shows the firstVisitMessage also when there are results but they are just loading still. Did you not notice? This would need a smarter way to test if there are really no items in the database. May be easier to finish this when the result loading indicator (#54) is done.

Chaitya62:
Hmm I think I ll figure out some other way to eliminate that !

Chaitya62:
Okay made some changes added a isFetching flag have added comments to why I set it to false and true everywhere. Initially it is true as we fetch the db whlle we load overview for the first time.I have added a loading placeholder which can be replaced by a loader later on.

src/overview/actions.js
```diff
@@ -30,6 +30,10 @@ export function refreshSearch() {
         const query = ourState(getState()).query
         const oldResult = ourState(getState()).searchResult
         filterVisitsByQuery({query}).then(searchResult => {
+
+            //set fetching results to true
+            ourState(getState()).isFetching = true;
```
obsidianart: this seems like direct state manipulation. is it the case?

Chaitya62:
Yes is it a bad practice?

src/overview/reducer.js
```diff
@@ -5,13 +5,16 @@ import * as actions from './actions'
 const defaultState = {
     searchResult: {rows: []},
     query: '',
+    isFetching: true,
```
obsidianart: why is default true?

src/overview/reducer.js
```diff
@@ -5,13 +5,16 @@ import * as actions from './actions'
...
 }
 
 function setQuery(state, {query}) {
+	state.isFetching = true; //will start fetching
```
obsidianart: I think the comment is incorrect, this doesn't trigger anything per se

src/overview/reducer.js
```diff
@@ -5,13 +5,16 @@ import * as actions from './actions'
...
     return {...state, query}
 }
 
 function setSearchResult(state, {searchResult}) {
+	state.isFetching = false; //will stop fetching
```
obsidianart: I think the comment is incorrect, this doesn't trigger anything per se

Chaitya62:
A the default is true because we are fetching from the database as soon as we start our app

src/overview/reducer.js
Chaitya62: What i mean was when we call setSearchResult means the database has returned a value now if the query is changed and we again make it true

src/overview/reducer.js
Chaitya62: Actaully I think instead of will start fetching. I should elaborate a little while intializing default

src/overview/reducer.js
obsidianart: I understand, I think the problem is around the comment. this is the state of the fetching, not something which trigger a behaviour.

obsidianart:
When you start fetching the flag is changed, the initial state is false regardless. The initial state is about what it is, and the fetching is trigger later on in the code. if it's not clear we can talk about it

Chaitya62:
Yes intuitively it should be false initially but then the message that only firstVisitors should see appears I guess isLoading is a better name then isFetching but as we discussed I should wait for the loader issue to be resolved

obsidianart:
wouldn't it be changed to true immediately (stating the search starts straight away)?

Chaitya62:
Yes but the component when default was false loads and goes away it look buggy where as when it's true initially the component loads with loading message

obsidianart:
ok good :)

Chaitya62:
How about I add 
```javascript 
export  isDatabaseEmpty = ()=>{
	let isEmpty = false;
	db.info().then((result)=>{
		if(result.doc_count === 0){
			isEmpty = true;
		}

	});
	return isEmpty;
}
```

in pouchdb.js ,to check database is empty or not? 
The above code won't work due to the async call but can we make a similar method in pouchdb to check the db is empty or not

Treora:
@Chaitya62: checking the database requires calling an asynchronous function, so it cannot be done while rendering and requires adding extra state. Seems too complicated to me. (your code always returns false by the way, promises don't work this way)

For simplicity, should we perhaps forget about the first-time message again (sorry, I know it was my own suggestion), and just check in the ResultList if there is a query but no results? Then we can think about the first-use message later again, maybe also as part of a more elaborate introduction.

Chaitya62:
@Treora just edited my previous message I realised that code won't work ! I ll just add the no result message and update the PR !

Chaitya62:
@Treora I think now the PR is ready ! please review it!

Treora:
Nice and simple, thanks! I made the message a bit more subtle in appearance, as it need not draw much attention. I like the fade-in, good idea!

Issue #50, opened 2017-03-05
Location changes by web-apps (through history API) are not logged as multiple pages/visits


rusuo:
1. Go to youtube (there will be one entry created in the history)
2. Select various songs, or leave the playlist to continue to the next song on its own

Actual results:
There is only one entry created (the initial one), no other entries under the same domain appear. 
Note: If I open a new tab then it will be created as a new entry.

Expected results:
A new entry is created for every different page visited

blackforestboi:
Also important to mention, that this only happens on a few sites, for example Youtube or GitHub or Twitter.

edit: I have a hunch that it may have something to do with web apps? Because if I visit pages like: https://shuttleworthfoundation.org it does not have a problem. Also useful to mention, that this also happens for the WorldBrain extension.

Treora:
Thanks for reporting! This is easily explainable, as the activity-logger currently [listens](https://github.com/WebMemex/webmemex-extension/blob/1423c9333112bf8f870764483bca2de14e250e9a/src/activity-logger/background/index.js#L22) to the [`browser.webNavigation.onCommitted`](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webNavigation/onCommitted) event. We should also start listening to [`onHistoryStateUpdated`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation/onHistoryStateUpdated) events to catch these location changes that web-apps perform (as Oliver rightly suspected). Possibly just three lines of code for now, though it needs to be tested, and we will probably have to handle such location changes somewhat differently later on (e.g. when deduplicating pages).

blackforestboi:
@Treora is this something that can be marked as newcomer issue?

gastonche:
I'll get work on this right away. So if i understand the task well, i'll have to listen for  onHistoryStateUpdated on the broswer and use it to add a new page to the extension.

mukeshkharita:
Hello,
I was trying to solve this issue
I added this code in activity-logger/background/index.js https://paste.ubuntu.com/25170020/
but this stores the link two times
so how can I resolve this problem?

Treora:
[...]

PR #59, opened 2017-03-08
Fixed issue on History no updating for web apps

gastonche:
PR pertaining to issue #50. Please do review soon and give me feedback. This is my first PR here so, please do point out any errors I may have made in the process, so i can correct it now and subsequently.

### DIFF
--- a/src/activity-logger/background/index.js
+++ b/src/activity-logger/background/index.js
@@ -3,6 +3,9 @@ import performPageAnalysis from '../../page-analysis/background'
 import storeVisit from './store-visit'
 import { isWorthRemembering } from '..'
 
+//array to store added pages for this section
+let pages = []
+
 // Create a visit/page pair in the database for the given URL.
 function logPageVisit({url}) {
     const timestamp = new Date()
@@ -18,8 +21,18 @@ function logPageVisit({url}) {
     return storeVisit({timestamp, visitInfo, pageInfo})
 }
 
-// Listen for navigations to log them and analyse the pages.
-browser.webNavigation.onCommitted.addListener(details => {
+let pageLoaded = function(item){
+    pages.forEach((page)=>{
+        if(item.url == page.url && item.tabId == page.tabId){
+
+            return true;
+        }
+    });
+    return false;
+}
+
+let navigated = function(details){
+     console.log(details);
     // Ignore pages loaded in frames, it is usually noise.
     if (details.frameId !== 0)
         return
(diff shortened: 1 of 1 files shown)


src/activity-logger/background/index.js
```diff
@@ -18,23 +21,57 @@ function logPageVisit({url}) {
...
     logPageVisit({
         url: details.url
-    }).then(
+    }).then(        
         // Wait until its DOM has loaded.
         value => whenPageDOMLoaded({tabId: details.tabId}).then(()=>value)
     ).then(({visitId, pageId}) => {
+
+        //add the loaded page to the array to prevent double loading due to double checks
+        pages.push({tabId: details.tabId, url: details.tabId});
```
Treora: `url: details.tabId` looks like a mistake.

src/activity-logger/background/index.js
```diff
@@ -18,23 +21,57 @@ function logPageVisit({url}) {
     return storeVisit({timestamp, visitInfo, pageInfo})
 }
 
-// Listen for navigations to log them and analyse the pages.
-browser.webNavigation.onCommitted.addListener(details => {
+let pageLoaded = function(item){
+    pages.forEach((page)=>{
+        if(item === page){
```
Treora: This tests strict object equality: given two objects whose properties are the same, it would still return false, unless they are really both references to the same object. Changing `===` to `==` will not change this either, you have to check for the equality of their properties (`tabId` and `url`)

src/activity-logger/background/index.js
```diff
@@ -18,23 +21,57 @@ function logPageVisit({url}) {
...
+// Listen for navigations to log them and analyse the pages.
+browser.webNavigation.onCommitted.addListener((details)=>navigated(details))
+
+// Listen for navigations in web apps to log them and analyse the pages.
+browser.webNavigation.onHistoryStateUpdated.addListener((details)=>{
+    browser.tabs.get(details.tabId,(tab)=>{
+        
+        //make sure the page logged is the nal page that is reached by the history change action
+        if(tab.url===details.url)
+            navigated(details)
```
Treora: Interesting that this is needed. In which cases did you find them to be different?

Treora (commented):
Thanks for trying this out, it looks like you discovered and worked around some complications. I see you made two workarounds, one that checks against double logging, and one that checks if the url is correct. I would like to understand why exactly those problems occur.

Keeping some state of which tabs are open (your 'pages' variable) is something we seem to need for many things. Probably we should take a more structural approach to this, instead of keeping an array. Maybe not in this PR though.

src/activity-logger/background/index.js
gastonche: I noticed in some cases, like youtube, when a video was clicked, the onHistoryStateChanges was fired twice, so i needed to check to make sure that only the right page was taken.

src/activity-logger/background/index.js
gastonche: yeah, that was an error, i'll fix that right away.

src/activity-logger/background/index.js
gastonche: okay, thanks for the correction. Let me fix that and submit a new PR.

gastonche:
I have updated the PR. You can check it now

Treora:
Thanks; I will have a look later and try it out myself.

Treora:
I tried it out and played around a bit, and see how you came to your solutions. I hope you don't mind I am not going to merge this PR however, as I would rather first revise the activity logger to have it keep some state about the tabs and their recent history, after which a solution like the one you found can be implemented quite elegantly. Thanks for exploring and reporting this!

gastonche:
Okay no qualms.

Issue #22, opened 2017-02-12
Deduplicate pages


Treora:
When the same page is visited multiple times, it is silly to create a new copy in the database for every visit. A non-trivial question however is when to consider two pages to be the same:

1. The easy case: if they have the same url and the same content (compare ETags or hashes), they are the same.
2. The easy opposite case: if they have different urls and different content, they are not the same.
3. The difficult case: if they have the same url, but the content has changed; we may want to consider them the same if they are very similar (e.g. only ads changed), but different if we consider a news site's front page. 
4. Also, the case nobody cares about: if they have different urls but the same content. Though it may not happen often enough to waste much space, there is an interesting aspect: spotting the sameness would allow e.g. reattaching the same notes. Probably something for later though.

Determining sameness can be done very crudely at first, and improved later to better detect the almost-the-same situations in case 3 above (and maybe case 4).

blackforestboi:
I can imagine that detecting content "sameness" can be quite difficult, if there is content on the page that is dynamic. 

How can we circumvent this? With news articles, I have the impression that readability does a quite decent job at cutting everything out, but not so much with other types of HTML content. 

I don't know if that idea already came up, but how about storing the text through readability in as priority 1 text and the diff between `body.content.innerText`- `readability_text` as priority 2 text? AFAIK pouchDB quick search lets you boost certain fields. 

I also liked the idea you had about creating content type (or even domain)-specific parser libraries. 

Another idea, that I believe came from @bwbroersma was to just store the diffs from each visit as pure text, means they are at least searchable. We could just append them to the `text` field.

bwbroersma:
Storing multiple visits of the same URL would be nice to detect [this](http://newsdiffs.org/diff/1104528/1105097/www.nytimes.com/2016/03/15/us/politics/bernie-sanders-amendments.html) kind of changes ([medium post about it](https://medium.com/@brokenravioli/proof-that-the-new-york-times-isn-t-feeling-the-bern-c74e1109cdf6)). Just storing (compressed) diffs, even of the complete HTML text should be quite minimal, if that is not doable do a (compressed) diff on the readability DOM (preferred over just the innerText), so you store ~~changes~~ edits in a meaningful way. Using a [preset deflate dictionary](https://blog.cloudflare.com/improving-compression-with-preset-deflate-dictionary/) per language would result in a minimal size diff, but remember this only becomes a problem when size becomes an issue, and a fix can be done retroactively with an update.

[...]

PR #60, opened 2017-03-09
Framework for page deduplication

Treora:
As discussed in issue #22.

A lot of code to do relatively little, but it provides the basic framework and a simple first implementation for checking whether two pages that were presumably retrieved from the same URL are the same document, completely different pages, or something in between (the more recent one presumably being an update to the other). A complex workaround to the lack of versioning in URLs.

Ideally we could tell directly whether a page is the same as last time or not, for example by checking the ETag HTTP header, if present. The code needed for such checks still has to be implemented, but the required boilerplate is included here (`tryReidentifyPage` in `src/page-storage/store-page.js`).

If a page was not reidentified beforehand, a new page doc is created in the database, and the usual page analysis is run on it. After this the analysed page contents are compared against the candidate page (= the previous page we got from the URL), currently by a simple text comparison between the `body.innerText` and title of the two pages (see `src/page-storage/sameness.js`). A level of 'sameness' is determined (rather subjectively), and this level is then used for deciding what to do with the two (in `src/page-storage/deduplication.js`). Currently the two possible actions are to either keep them both as-is, or to forget the analysed contents of the older page and add a `seeInstead` field that redirects any reader to the new page; much like HTTP redirects. Which actions are taken for which sameness levels can be worked out in more detail later.

If anybody would like to share their thoughts on the sanity of the direction of this approach, I would be glad to hear. There are some important design choices embedded in here. Regard this code as a first iteration though, I hope the approach will flesh out and evolve further over time.


src/page-storage/sameness.js
```diff
@@ -0,0 +1,68 @@
...
+// Tell how similar two strings are in a qualitative way.
+function textSameness(text1, text2) {
+    if (!text1 || !text2)
+        return Sameness.UNKNOWN
+
+    if (text1 === text2)
+        return Sameness.EXACTLY
+
+    const normaliseWhitespace = s => s.trim().replace(/s+/g, ' ')
+    if (normaliseWhitespace(text1) === normaliseWhitespace(text2))
```
obsidianart, line 35: I'm surprised this check is needed and you don't get something like 0.99 in this case

obsidianart:
It seems indeed a lot of code to do one thing. It might be worth doing it for future development but as for today my impression is that 50% of it is currently unused and you just end up either replacing a page or not. What is the final picture of this approach? What do you want to give the user?

src/page-storage/sameness.js
Treora: The idea is that 'ostensibly' means that any changes are of no visible influence, or at least of no semantic influence. A single letter is considered a tiny but actual change to the content, but an extra space is regarded unimportant, and in html it would normally not even appear on screen. Also for other data types, 'ostensibly the same' would mean that the rendered result is the same.

In any case, the tests here are largely intended as simple examples, to be worked out in more detail and for more data types another time.

Treora:
> What is the final picture of this approach? What do you want to give the user?

I want to bring versioning to the web, and provide users with a model of webpages that matches the way people think about it. The view that one URL identifies one document is simply wrong. The current approach, that each time you dereference a URL you get a completely unrelated document, swings too far to the other side. We need something in between.

Today's newspaper frontpage is a different document than yesterday's, though they share the same URL. Today's view of a specific article may be still be the same article however, even though the advertisements around it changed. And tomorrow the article may have had a small correction, making it a revision of the same document.

obsidianart:
ok, it's a similar approach to the internet archive. Good idea.

Issue #53, opened 2017-03-05
Show a message for "no results found"


rusuo:
When searching does not return any result, show a message to inform the user that no results were found

Chaitya62:
Submitting a PR soon

blackforestboi:
What is the progress of this? @Chaitya62

rachanamamillapalli:
Can I work on this task? - newcomer

oliversauter:
@rachanamamillapalli already in the making, as the label suggests :)
 
Here is the corresponding PR: https://github.com/WebMemex/webmemex-extension/pull/56

Chaitya62:
@oliversauter. PR review is left

davidemdot:
Hi, @oliversauter,

Last message means this issue is free now?

Chaitya62:
@davidemdot no I am working on it

davidemdot:
Sorry, @Chaitya62, I misunderstood you.

PR #61, opened 2017-03-10
fixed Show a message for no results found #53

Rohanhacker:
screenshot from 2017-03-10 19-33-24


Rohanhacker:
this is my first commit here

Treora:
Thanks! However there's also #56 tackling exactly the same issue! It is just delayed a bit because we don't want to have a different message when the database is simply empty. Closing this one.

Issue #15, opened 2017-02-11
Wrong page is captured


Treora:
If the tab's location changes before loading of some page completes, the next page's load event will trigger the analysis of that former page, which then thus gets the wrong page's title, screenshot, etcetera.

Should be fixable in `tab-events` (see 043be83).

gastonche:
Please i would like to know if this issue is still a problem, so that i may work on it. Thanks

Treora:
Still to be fixed indeed, thanks for getting at it. Not sure exactly which event to listen to to be 100% sure the promise is cancelled when the page is changed, so have a look first what options are available.

gastonche:
okay i'll try various options until we get to a good fit

Treora:
Fixed by #62.

blackforestboi:
Congrats @gastonche to your first Merged PR :)

gastonche:
Thanks @oliversauter

PR #62, opened 2017-03-11
[background] added events to ensure right page capture

gastonche:
PR pertaining to #15 .
I have added event listeners to all tab watching events that ensure if the tab url changes , it will reject the promise such that the watching action can't continue

### DIFF
--- a/src/util/event-to-promise.js
+++ b/src/util/event-to-promise.js
@@ -16,10 +16,10 @@ export default function eventToPromise({
     reject: rejectOpts,
 }) {
     // We can get a single options object, or an array of them, or undefined.
-    if (typeof resolveOpts !== 'Array') {
+    if (!Array.isArray(resolveOpts)) {
         resolveOpts = (resolveOpts !== undefined) ? [resolveOpts] : []
     }
-    if (typeof rejectOpts !== 'Array') {
+    if (!Array.isArray(rejectOpts)) {
         rejectOpts = (rejectOpts !== undefined) ? [rejectOpts] : []
     }
 
--- a/src/util/tab-events.js
+++ b/src/util/tab-events.js
@@ -1,40 +1,44 @@
 import eventToPromise from './event-to-promise'
 
+const tabChangedEvents = [
+    {
+        event: browser.webNavigation.onCommitted,
+        filter: (details)=>(details.tabId == tabId && details.frameId == 0),
+        reason: {message: "Tab URL changed before event occurred."}
+    },
+    {
+        event: browser.webNavigation.onHistoryStateUpdated,
+        filter: (details)=>(details.tabId == tabId && details.frameId == 0),
+        reason: {message: "Tab URL changed before event occurred."}
+    },
+    {
+        event: browser.tabs.onRemoved,
+        filter: (closedTabId) => (closedTabId === tabId),
+        reason: {message: "Tab was closed before event occurred."},
+    }
+];
+
+
 // Resolve if or when the page DOM is loaded (document.readyState==='interactive')
 // Rejects if it is closed before that.
 // XXX Needs host permission on the tab
-// TODO reject if loading of the current page is aborted. (#15)
 export function whenPageDOMLoaded({tabId}) {
-    // This more obvious approach can get stuck in limbo, as there is no
-    // tab.status==='interactive'; it is either 'loading' or 'complete'.
-    // return browser.tabs.get(tabId).then(tab => {
-    //     if (tab.status === 'complete') // XXX Tests the wrong thing.
-    //         return // Resolve directly
-    //
-    //     return eventToPromise({
-    //         resolve: {
-    //             event: browser.webNavigation.onDOMContentLoaded,
-    //             filter: ({tabId: eventTabId}) => (eventTabId===tabId),
-    //         },
-    //         reject: {
-    //             event: browser.tabs.onRemoved,
-    //             filter: (closedTabId) => (closedTabId === tabId),
-    //             reason: {message: "Tab was closed before DOM was loaded."},
-    //         },
-    //     })
-    // })
-
-    // Workaround: we run a script at 'document_end' (= 'interactive')
-    return browser.tabs.executeScript(tabId, {
-        code: 'undefined', // Bogus code, anything works.
-        runAt: 'document_end',
-    })
+    
+    return new Promise((resolve, reject) => {
+        // using execute script here as a simple approach since there is no tab.status==='interactive'
+        browser.tabs.executeScript(tabId, {
+            code: 'undefined',
+            runAt: 'document_end',
+        }).then(() => resolve());
+        
+        eventToPromise({
+            reject: tabChangedEvents
+        }).catch(reject);
+    });
 }
-
 
 // Resolve if or when the page is completely loaded.
 // Rejects if it is closed before that.
-// TODO reject if loading of the current page is aborted. (#15)
 export function whenPageLoadComplete({tabId}) {
     return browser.tabs.get(tabId).then(tab => {
         if (tab.status === 'complete')
@@ -46,11 +50,7 @@ export function whenPageLoadComplete({tabId}) {
                 filter: (changedTabId, {status}) =>
                     (changedTabId === tabId && status === 'complete'),
             },
-            reject: {
-                event: browser.tabs.onRemoved,
-                filter: (closedTabId) => (closedTabId === tabId),
-                reason: {message: "Tab was closed before loading was complete."},
-            },
+            reject: tabChangedEvents
         })
     })
 }
@@ -58,7 +58,6 @@ export function whenPageLoadComplete({tabId}) {
 
 // Resolve if or when the tab is active.
 // Rejects if it is closed before that.
-// TODO reject (optionally) if the tab's url changes (='whenPageActive'?) (#15)
 export function whenTabActive({tabId}) {
     return browser.tabs.query({active:true}).then(
         activeTabs => (activeTabs.map(t=>t.id).indexOf(tabId) > -1)
@@ -71,11 +70,7 @@ export function whenTabActive({tabId}) {
                 event: browser.tabs.onActivated,
                 filter: ({tabId: activatedTabId}) => (activatedTabId === tabId),
             },
-            reject: {
-                event: browser.tabs.onRemoved,
-                filter: (closedTabId) => (closedTabId === tabId),
-                reason: {message: "Tab was closed before it became active."},
-            },
+            reject: tabChangedEvents
         })
     })
 }


src/util/tab-events.js
```diff
@@ -1,5 +1,20 @@
 import eventToPromise from './event-to-promise'
 
+function tabChanged(tabId) {
+    return new Promise((resolve, reject) => {
+        browser.webNavigation.onCommitted.addListener((details)=>{
+            console.log(details);
```
obsidianart: any reason to leave a console.log here?

src/util/tab-events.js
```diff
@@ -1,5 +1,20 @@
...
+            if(details.tabId == tabId && details.frameId == 0)
+                reject();
+        });
+        browser.webNavigation.onHistoryStateUpdated.addListener((details)=>{
+            console.log(details);
```
obsidianart: any reason to leave a console.log here?

src/util/tab-events.js
```diff
@@ -1,5 +1,20 @@
 import eventToPromise from './event-to-promise'
 
+function tabChanged(tabId) {
+    return new Promise((resolve, reject) => {
+        browser.webNavigation.onCommitted.addListener((details)=>{
+            console.log(details);
+            if(details.tabId == tabId && details.frameId == 0)
+                reject();
```
obsidianart: this might cause a double rejection since line 13 does the same thing. Even if this is allowed by the A+ pattern (the second rejection would be ignored) I wonder if there's any point in having a listener working but the promise rejected this way.

src/util/tab-events.js
```diff
@@ -23,12 +38,15 @@ export function whenPageDOMLoaded({tabId}) {
...
-
-    // Workaround: we run a script at 'document_end' (= 'interactive')
-    return browser.tabs.executeScript(tabId, {
-        code: 'undefined', // Bogus code, anything works.
-        runAt: 'document_end',
-    })
+    return new Promise((resolve, reject) => {
+        // Workaround: we run a script at 'document_end' (= 'interactive')
+        browser.tabs.executeScript(tabId, {
+            code: 'undefined', // Bogus code, anything works.
```
obsidianart: if you say```code: 'boguse_code_anything_works'``` you don't need the comment

src/util/tab-events.js
```diff
@@ -23,12 +38,15 @@ export function whenPageDOMLoaded({tabId}) {
...
+            runAt: 'document_end',
+        }).then(resolve());
```
obsidianart: this is incorrect and would call too early since you are invoking at declaration time, this should be
```
then(resolve) //if you want to pass the value of the then
```

or

```
.then(()=>resolve()) //if you want to resolve with empty value
```

obsidianart: Also, there's no catch and this is discourage (and will be disallowed in node soon)

src/util/tab-events.js
```diff
@@ -23,12 +38,15 @@ export function whenPageDOMLoaded({tabId}) {
...
+        
+        tabChanged(tabId).catch(() => reject());
```
obsidianart: why is this needed at all?

gastonche: This is needed, in order to check if the tab changes, so it rejects

src/util/tab-events.js
```diff
@@ -40,18 +58,24 @@ export function whenPageLoadComplete({tabId}) {
...
-                    (changedTabId === tabId && status === 'complete'),
-            },
-            reject: {
-                event: browser.tabs.onRemoved,
-                filter: (closedTabId) => (closedTabId === tabId),
-                reason: {message: "Tab was closed before loading was complete."},
-            },
-        })
+        return new Promise((resolve, reject) => {
+            eventToPromise({
```
obsidianart: @Treora I'll leave you evaluate if this library is correct or not here

src/util/tab-events.js
```diff
@@ -40,18 +58,24 @@ export function whenPageLoadComplete({tabId}) {
...
+                    filter: (changedTabId, {status}) =>
+                        (changedTabId === tabId && status === 'complete'),
+                },
+                reject: {
+                    event: browser.tabs.onRemoved,
+                    filter: (closedTabId) => (closedTabId === tabId),
+                    reason: {message: "Tab was closed before loading was complete."},
+                },
+            }).then(()=>resolve())
+            .catch(()=>reject());
```
obsidianart: this should probably be ```.catch(reject)``` otherwise the reject reason will be lost making quite hard to debug

src/util/tab-events.js
```diff
@@ -40,18 +58,24 @@ export function whenPageLoadComplete({tabId}) {
...
+
+        tabChanged(tabId).catch(() => reject());
```
obsidianart: why is this needed at all?

src/util/tab-events.js
gastonche: okay

src/util/tab-events.js
obsidianart: ok, interesting pattern, I'll leave it to @Treora  :)

gastonche:
I will fix the corrections pointed out in the reviews, and update the PR,
I added the tabChanged method which returns a promise that rejects if the tab is changed, that way, the function waiting for a resolve will stop. In that way we  can avoid taking the snapshot of the wrong page
@obsidianart

obsidianart:
@gastonche The approach is clear, but I've never seen promises used for an event based pattern so I'd leave Treora decide

src/util/tab-events.js
gastonche: My view on this was that, i have to listen to changes in the tabs url coming both from normal redirects like links clicked, and from changes triggered by the HTML 5 History API,  and reject in both cases, but i doubt there is ever a case  where those two can resolve at once and so this can only call reject once
@obsidianart

src/util/tab-events.js
gastonche: Terrible error

Treora:
@obsidianart: It may not be very common, but promises are quite a fitting abstraction when waiting for an event to happen exactly once; especially when there is the need to stop waiting when another type of event happens. I wrote this small `eventToPromise` helper exactly because I ended up using this pattern several times.

@gastonche: looks like you got a technically sensible solution, roughly what I had in mind. I have a question the choice of events to listen to, and I think we can refactor the code to make it more understandable to a reader. Will comment inline.

gastonche:
@Treora It would be nice to have it clearer for sure. What do you have in mind

src/util/tab-events.js
```diff
@@ -1,57 +1,59 @@
...
+        browser.webNavigation.onCommitted.addListener((details)=>{
+            if(details.tabId == tabId && details.frameId == 0)
+                reject();
+        });
+        browser.webNavigation.onHistoryStateUpdated.addListener((details)=>{
+            if(details.tabId == tabId && details.frameId == 0)
+                reject();
+        });
+    });
+}
```
Treora: My question about the choice of events: why do you choose to listen to onCommited and onHistoryStateUpdated? Is it ensured that these will capture all cases where a tab 'unloads' its page? There's a few more options in the [`webNavigation` API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/webNavigation), and also listening to `tabs.onChange` to check for URL changes would be an option that comes to mind. Did you try or research any of these options?

I guess you may have chosen these because the activity logger also listens to them (when #50 is resolved). Listening to the same events could then indeed solve the problem of accidentally capturing the wrong page, so that's a smart solution. If it is possible it would be nice however to ensure we have a watertight solution, that could also be trusted to work for potential other scenarios.

src/util/tab-events.js
```diff
@@ -1,57 +1,59 @@
 import eventToPromise from './event-to-promise'
 
+function tabChanged(tabId) {
+    return new Promise((resolve, reject) => {
+        browser.webNavigation.onCommitted.addListener((details)=>{
+            if(details.tabId == tabId && details.frameId == 0)
+                reject();
```
Treora: Instead of rejecting, it would be more logical to resolve here, since the name is tabChanged (I'd call it `whenTabChanged` for consistency, or an even more accurate description). Where it is used code would then switch from `catch` to `then`.

src/util/tab-events.js
Treora: I think I made eventToPromise such that it accepts multiple event descriptions to listen for; instead of creating a new Promise each time, could the events defined in `tabChanged` be described in an array that can be added to the `reject` argument of eventToPromise instead?

I am thinking of something like this:
```
const tabChangedEvents = [
    {
        event: browser.tabs.onCommit
        filter: details => (details.tabId == tabId && details.frameId == 0)
    },
    .... same for onHistoryStateUpdated or other events ...
]
```

src/util/tab-events.js
```diff
@@ -1,57 +1,59 @@
...
-                event: browser.tabs.onUpdated,
-                filter: (changedTabId, {status}) =>
-                    (changedTabId === tabId && status === 'complete'),
-            },
-            reject: {
-                event: browser.tabs.onRemoved,
-                filter: (closedTabId) => (closedTabId === tabId),
-                reason: {message: "Tab was closed before loading was complete."},
-            },
-        })
```
Treora: So here the value of `reject` would be an array:
```
reject: [{
    event: browser.tabs.onRemoved,
    filter: ...,
    reason: ...,
    },
   ...tabChangedEvents,
]
```

src/util/tab-events.js
Treora: (forget about this comment, if we replace this `tabChanged` Promise with a list)

gastonche:
Okay @Treora , i'll try to refactor the Code in order to use the event to promise approach

gastonche:
@Treora I have refactored the code to use the eventToPromise Approach. But then I left the new Promise in the whenPageDOMLoaded method because the browser.tabs.executeScript could not be added to the list of tabsChangedEvents since it dows not have  an addlistener property and so will throw an error. I believe it's more readable this way.

obsidianart:
@Treora I think event to Promise is a way to force a pattern within another, especially since you are waiting for rejection and not fullfilment. A promise rejection can come from many things (including syntax error).
Comparing the following I find this approach quite singular, also considering Node has events emitter (https://nodejs.org/api/events.html).
It might be my limit, but I always saw promises as a way to ask for something which you expect to happen, not to throw for something which you don't expect to happen.

```
//Promise
const waitForSmt = smt() 
waitForSmt.catch(doAction)
```

```
//Event
const waitFor = smt()
waitFor.on('smt', doAction)
```

```
//Callback
const waitForSmt = smt(doAction)
```

src/util/event-to-promise.js
```diff
@@ -16,10 +16,10 @@ export default function eventToPromise({
     reject: rejectOpts,
 }) {
     // We can get a single options object, or an array of them, or undefined.
-    if (typeof resolveOpts !== 'Array') {
+    if (!Array.isArray(resolveOpts)) {
```
Treora, line 19: Thanks for fixing, my bad! I wrote this code too hastily.

src/util/tab-events.js
```diff
@@ -1,56 +1,55 @@
...
 // Resolve if or when the page is completely loaded.
 // Rejects if it is closed before that.
-// TODO reject if loading of the current page is aborted. (#15)
 export function whenPageLoadComplete({tabId}) {
     return browser.tabs.get(tabId).then(tab => {
         if (tab.status === 'complete')
             return // Resolve directly
 
-        return eventToPromise({
+    return eventToPromise({
```
Treora: indent

src/util/tab-events.js
```diff
@@ -65,17 +64,12 @@ export function whenTabActive({tabId}) {
     ).then(isActive => {
         if (isActive)
             return // Resolve directly
-
         return eventToPromise({
```
Treora: lost a line here

src/util/tab-events.js
```diff
@@ -1,56 +1,55 @@
...
+    },
+    {
+        event: browser.webNavigation.onHistoryStateUpdated,
+        filter: (details)=>(details.tabId == tabId && details.frameId == 0),
+        reason: "Tab URL changed before loading was complete"
+    },
+    {
+        event: browser.tabs.onRemoved,
+        filter: (closedTabId) => (closedTabId === tabId),
+        reason: {message: "Tab was closed before loading was complete."},
```
Treora: These messages were different for each function, "before loading was complete" is a confusingly wrong message when printed by the whenTabActivated function. We can make them generic enough to match every of the functions, e.g. ".... before event occurred."
Also, it should be an object with a message field, not a string like you used above.

src/util/tab-events.js
```diff
@@ -1,56 +1,55 @@
...
-    //             event: browser.webNavigation.onDOMContentLoaded,
-    //             filter: ({tabId: eventTabId}) => (eventTabId===tabId),
-    //         },
-    //         reject: {
-    //             event: browser.tabs.onRemoved,
-    //             filter: (closedTabId) => (closedTabId === tabId),
-    //             reason: {message: "Tab was closed before DOM was loaded."},
-    //         },
-    //     })
-    // })
```
Treora, line 25: Removing the bulk of this commented code is fine, but I would still add an explanation why we use this `executeScript` approach: there is no tab.status==='interactive'.

Treora (commented):
Looks pretty, at least within the given approach of using eventToPromise, which @obsidianart made me doubt about (let's first merge this before reconsidering the approach though).

Comments inline. And another small thing: forgot to remove the TODO at `whenTabActive`. Also we could perhaps rename it now to `whenPageActive`, since it is not about the tab in general anymore (we could have an argument determine whether to wait for the specific page or the tab in general, but let's do that another time, if we will need it).

Treora:
> I think event to Promise is a way to force a pattern within another, especially since you are waiting for rejection and not fullfilment

@obsidianart: We are waiting for fulfilment. E.g. `whenPageLoaded` resolves when the page completed loading. It rejects if the page was closed before that ever happened.

I do understand and agree with your point that exceptions/rejections should be used for error handling, rather than for handling events that could be expected to happen. I am not sure if a page closing before it being loaded can reasonably be regarded as an exception.. For the user of the promise however, it probably does not matter if the promise rejected because of some really unforseeable error or because the page closed; either way what they waited for is not happening anymore.

I will think whether some event system can be more appropriate here, while still retaining the simplicity in usage, but I feel this promise approach is still acceptable.

obsidianart:
Feel free to merge, I have no truth.

Treora:
Great, looks ready to merge.
@gastonche: would you be okay with waiving your copyrights on your contributions and publish it in the public domain?

gastonche:
no problem @Treora . I am more interested in seeing the fixes in the repo

Issue #14, opened 2017-02-10
Search memory in address bar


Treora:
Two possible approaches:

1. Use the [`omnibox`](https://developer.chrome.com/extensions/omnibox) API, like [WorldBrain](https://github.com/WorldBrain/Research-Engine/)/[Falcon](https://github.com/lengstrom/falcon). Currently only possible in Chrome|ium, and only works when the user enters a specified prefix, but it does allow to directly display results as suggestions.
2. Define a search engine provider. If extension URLs are not supported as search providers, we can provide a fake URL, requests to which we redirect back to the extension. Downside: suggestions would probably not be supported as they seem not redirectable from the webRequest API. Also, switching default search engines is a hassle in Chrome|ium, though easier in Firefox.

Neither solution is really satisfying.

blackforestboi:
Another option would be to make it like Pinboard. So if users search via the address bar they automatically land on a search engine (maybe google) and then get (the first 5) results displayed on the side or as the first results. 

Also the missing API endpoint for the address bar will be resolved with the new FF53 (release 23. april). The problem then still remains for other browsers like safari or vivaldi.

Treora:
Indeed an option to peg onto a user's existing web search (though, again, not really satisfying).

None of these options exclude each other, and all of them are compromises just to provide somewhat more convenient, quicker access to something that is already accessible in another way. I guess some user research may be helpful here, and whatever works suffices in the meantime.

blackforestboi:
I think its best to just go with the way Falcon/WorldBrain does it right now. Accessing search it in the Omnibar with the command w+tab/space.

The code, giving inspriation for that can be found here: https://github.com/WorldBrain/Research-Engine/blob/master/src/js/background.js#L17
The functions involved: 
 - chrome.omnibox.onInputChanged.addListener(omnibarHandler);
 - chrome.omnibox.onInputEntered.addListener(acceptInput);
 - omnibarHandler
 - suggestionsComplete
 - makeSuggestions
 - dispatchSuggestions

## Addition to the existing code
The first result should be a link to the full results page of the same query.

Rohanhacker:
can I work on this I am a newcomer ?

blackforestboi:
Yes, good one. As discussed in the call, feel free to take it on!

Rohanhacker:
How does it look?
[...]

PR #63, opened 2017-03-12
Address issue -> Search memory in address bar #14

Rohanhacker:
screenshot from 2017-03-12 10-33-38

#63

### DIFF
--- a/src/background.js
+++ b/src/background.js
@@ -1,4 +1,51 @@
 import './activity-logger/background'
+import { filterVisitsByQuery } from './search'
+import niceTime from './util/nice-time'
+
+const makeSuggestion = (query, suggest) => {
+    filterVisitsByQuery({query, limit: 5}).then(searchResult => {
+        const suggestions = []
+        if(searchResult.rows.length === 0) {
+            chrome.omnibox.setDefaultSuggestion({description: 'No results found'})
+        } else {
+            chrome.omnibox.setDefaultSuggestion({description: 'Select an option below'})
+        }
+        searchResult.rows.map((result) => {
+            let doc = result.doc
+            let content = doc.url
+            let title = ''
+            let visitDate = niceTime(doc.visitStart)
+            let url = "<url>" + (content.slice(0,30)) + "</url>"
+            if(url.length > 30) {
+                url+= '...'
+            }
+            if(doc.page.extractedMetadata && doc.page.extractedMetadata.title) {
+                title = "<match>" + escape(doc.page.extractedMetadata.title) + "</match>"
+                title = title.replace(/%20|%2C|%3A|%3F|%B7|%23|%26/gi, matched => ' ')
+            }
+            let description = `${url} ${(visitDate)}  ${title}`
+            suggestions.push({
+                content,
+                description
+            })
+        })
+        suggest(suggestions)
+    })
+}
+
+const acceptInput = (text, disposition) => {
+    switch (disposition) {
+    case 'currentTab':
+        chrome.tabs.update({url: text})
+        break
+    case 'newForegroundTab':
+        chrome.tabs.create({url: text})
+        break
+    case 'newBackgroundTab':
+        chrome.tabs.create({url: text, active: false})
+        break
+    }
+}
 
 function openOverview() {
     browser.tabs.create({
@@ -16,3 +63,9 @@ browser.commands.onCommand.addListener(command => {
         openOverview()
     }
 })
+
+// This event is fired with the user change the input in the omnibox.
+chrome.omnibox.onInputChanged.addListener(makeSuggestion)
+
+// This event is fired with the user accepts the input in the omnibox.
+chrome.omnibox.onInputEntered.addListener(acceptInput)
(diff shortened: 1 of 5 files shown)


src/background.js
```diff
@@ -16,3 +63,9 @@ browser.commands.onCommand.addListener(command => {
         openOverview()
     }
 })
+
+// This event is fired with the user change the input in the omnibox.
+chrome.omnibox.onInputChanged.addListener(makeSuggestion)
+
+// This event is fired with the user accepts the input in the omnibox.
+chrome.omnibox.onInputEntered.addListener(acceptInput)
```
Treora: `chrome` should be replaced with `browser` throughout the file. We pursue a cross-browser approach, and use [webextension-polyfill](https://github.com/mozilla/webextension-polyfill/) to harmonise the APIs of Chrome|ium and perhaps others. Easiest is to follow [MDN's docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/omnibox/) while developing, it also contains info about browser compatibility.

src/background.js
```diff
@@ -1,4 +1,51 @@
 import './activity-logger/background'
+import { filterVisitsByQuery } from './search'
+import niceTime from './util/nice-time'
+
+const makeSuggestion = (query, suggest) => {
```
Treora: Perhaps start with a `setDefaultSuggestion` to tell search is busy?

src/background.js
```diff
@@ -1,4 +1,51 @@
...
+    filterVisitsByQuery({query, limit: 5}).then(searchResult => {
+        const suggestions = []
+        if(searchResult.rows.length === 0) {
+            chrome.omnibox.setDefaultSuggestion({description: 'No results found'})
+        } else {
+            chrome.omnibox.setDefaultSuggestion({description: 'Select an option below'})
```
Treora: Sounds obvious. Is a default suggestion required, or could it just be removed when results arrive? Or perhaps something like "Found these pages in your memory:" would be helpful to explain where things came from.

src/background.js
```diff
@@ -1,4 +1,51 @@
...
+        }
+        searchResult.rows.map((result) => {
+            let doc = result.doc
+            let content = doc.url
+            let title = ''
+            let visitDate = niceTime(doc.visitStart)
+            let url = "<url>" + (content.slice(0,30)) + "</url>"
+            if(url.length > 30) {
```
Treora: Testing the wrong variable. Did you mean `content.length`?

src/background.js
```diff
@@ -1,4 +1,51 @@
...
+                url+= '...'
+            }
+            if(doc.page.extractedMetadata && doc.page.extractedMetadata.title) {
```
Treora: You could use `page.title`, which is set [here](https://github.com/WebMemex/webmemex-extension/blob/06bf788a001699ef51a89e7d469c3546d8c4a022/src/page-analysis/index.js#L17-L19) to whatever title is available. If no title is known at all (e.g. the page never loaded), it is set to the url though, which is a bit ugly; but in that case it will probably not show up in the search results either.

src/background.js
```diff
@@ -1,4 +1,51 @@
...
+                title = "<match>" + escape(doc.page.extractedMetadata.title) + "</match>"
+                title = title.replace(/%20|%2C|%3A|%3F|%B7|%23|%26/gi, matched => ' ')
```
Treora: What is the idea here? When do titles contain these encoded characters?

src/background.js
```diff
@@ -1,4 +1,51 @@
...
+            }
+            let description = `${url} ${(visitDate)}  ${title}`
+            suggestions.push({
```
Treora: You are performing a map, so no need to push elements: `const suggestions = searchResult.rows.map(....)`
Also, the function within map could be factored out and called `visitToSuggestion` or something understandable.

Treora (commented):
Great, looks neat. I made some comments in-line. One more thing: to keep things clean and modular, you could put all your code in a new file (call it omnibar, or omnibar-suggestions or something) in `src/`, and import that file from `src/background.js`

src/background.js
blackforestboi: Good idea, with that we can mitigate a slow search and people get feedback that something is happening.

src/background.js
Rohanhacker: escape replaces -, #, spaces with %20, %2C etc so it replaces them with spaces otherwise It'll look ugly.

Treora:
@Rohanhacker: I did that bunch of edits myself as it seemed quicker than commenting on each aspect, I hope you agree with the changes. Unless you see problems still, I think this is ready to merge. The remaining to-do or other improvements can of course be resolved later.

The usual question for legal clarity: are you willing to waive all your copyrights on this code, to dedicate it to the public domain? (see e.g. unlicense.org)

Rohanhacker:
yup certainly.
I dedicate any and all copyright interest in this software to the
public domain. I make this dedication for the benefit of the public at
large and to the detriment of my heirs and successors. I intend this
dedication to be an overt act of relinquishment in perpetuity of all
present and future rights to this software under copyright law.

Treora:
Merged. Thanks!

Issue #20, opened 2017-02-12
Editable blacklist/'forgetlist'


Treora:
As a user, I do not want my computer to remember visits to, ehm, well, certain websites.

The settings page should have an editable list of domain names which are to be ignored. Each entry could be a regular expression to provide more power to power users.

I noticed the word 'blacklist' appeared a bit confusing to some people (we do not block the pages). Perhaps call it 'forget-list', 'ignore-list', or something else?

blackforestboi:
I like the wording "Ignore-List"

Question we may consider: Should we automatically delete all pages that have been blacklisted in retrospect as well?

Treora:
> Should we automatically delete all pages that have been blacklisted in retrospect as well?

Good point. Would be good to make that possible, but it would preferably first ask whether to "delete 62 matching pages from the memory". Else I would probably erase half my memory some day through a silly regex mistake. ;)

blackforestboi:
> but it would preferably first ask whether to "delete 62 matching pages from the memory"

Good idea. Maybe just with a simple browser prompt that can be cancelled? 

![](https://i.stack.imgur.com/gyxZZ.png)

Treora:
Something like that, I suppose. Though preferably with something slightly prettier (plenty of UI libraries provide question dialogs). It could also be less obtrusive by the way, I now imagine a small note below the list asking "Also remove 62 matching items from memory? [Yes] [No]". Sorry for fussing over details already.

blackforestboi:
hehe yeah sure :) I am not thinking about that specific UI, was just as an example. 

> "Also remove 62 matching items from memory? [Yes] [No]"

Sounds good to me. Simple short and to the point.

blackforestboi:
@aquibm 
Do you plan to take on the UI and the logic part of this task?
With the logic part I mean the actual storage of the blacklist items to any form of storage.

If yes, we need some clarification on how you aim to store the items, an info needed for another task: The actual enforcement of the blacklist. 
This would be a function that tests a URL if it should be stored or not. For that we need to know how the list of items to check is loaded. 

Just that its not left unsaid: Some options we talked about in the past would to simply store it in local storage, or to also store it to PouchDB. The latter option would have the advantage, that we can replicate the blacklist settings as well (without additonal work), when the DB is replicated (i.e. on the server or to another installation)
Though this part does not solve the question of the format.

aquibm:
[...]

PR #65, opened 2017-03-12
UI for history import (fix #17)

### LINKED ISSUE #17: UI for bookmark&history import
A user interface to start and stop importing of bookmarks & history, and watch its progress.

See description here: https://github.com/WorldBrain/Research-Engine/issues/92

alxmiron:
### Functionality 
Since this PR is about UI only, I didn't change/implement any loading or parsing logic as it can be serious work in separate issue. Here are just implemented components and containers, that get statistics (showed within table) from props and dispatches actions.

When "Start import" button is clicked, `startLoadingHistory` action is dispatched.
While loading, user can click "Pause", that dispatches `stopLoadingHistory` action with status `paused`.
After that user can resume import by "Resume" button, that dispatches `resumeLoadingHistory` action.

Bookmarks loading and action handlers for it were not implemented as [bookmarks import functionality not finished](https://github.com/WebMemex/webmemex-extension/issues/29)

### Please, review all comments about the code

------------------
I have some global suggestion to the whole project. I don't know where to write them, I'll leave them here

### Suggesstions
- add eslint (! important, to standardize the code and prevent spelling mistakes)

- file structure is complicated for me - core functionality mixed with view components and containers, actions are inside components etc. My suggestion for file structure:
modules/
__activity-logger/
__browser-history/
__search/
__page-analisys/
__overview/  - actions & epics & reducer - main logic
__options/     - actions & epics & reducer - main logic
components/  - pure react components (stateless if possible) + specific styles for it
__overview/  
__options/
__page-viewer/
containers/  - containers (mapStateToProps and mapDispatchToProps connections only)
__overview/
__options/
store/  - store creation, exporting
styles/  - global styling, styling for common components (buttons etc)
util/
dev/  - redux devtools creation, exporting

- why is store creation inside overview folder? it relates to the whole app, not to overview page only
- same about Redux Devtools setup
- set base path to src/ to avoid '../../../../../components' problem
- move src/overview/readme.md with develop guidelines to src root folder
- remove key "applications" from manifest file for Chrome extension (it's for FF, Chrome doesn't recognize it)

Will be glad to discuss all written above
#17


src/browser-history/background/import-history.js
```diff
@@ -116,32 +117,66 @@ export default function importHistory({
...
-    })
+    return function(dispatch, getState) {
+        dispatch(startLoadingHistory())
+        // Get the full history: both the historyItems and visitItems.
+        console.time('import history')
+        return getHistoryItems({startTime, endTime})
+            .then(data => {
+                if (getState().options.loadingHistoryStatus != 'pending') {
+                    // console.log('stopped on step 1')
+                    throw { status: 'canceled' }
```
alxmiron: The only way I found yet how to stop Promise execution: check whether stop button was clicked after each async action within a promise

Loading logic wasn't changed in this file

src/options/components/settings/index.jsx
```diff
@@ -0,0 +1,12 @@
...
+
+
+const SettingsComponent = () => (
+    <div>
+        <h1 className={styles.routeTitle}>Settings</h1>
+        <p>Settings specific components to come in here.</p>
+    </div>
+)
+
+export default SettingsComponent
```
alxmiron, line 12: Settings page wasn't changed. I just split it into component and container and located in correct folders

src/options/containers/import/index.jsx
```diff
@@ -1,17 +1,27 @@
...
-        return (
-            <div>
-                <h1 className={routeTitle}>Import</h1>
-                <p>Import specific components to come in here.</p>
-            </div>
-        )
+const mapStateToProps = state => ({
+    loadingHistoryStatus: ourState(state).loadingHistoryStatus,
+    searchIndexRebuildingStatus: ourState(state).searchIndexRebuildingStatus
+})
```
alxmiron: We should use stateless components instead of `React.Component` where it's possible ([more info](https://hackernoon.com/react-stateless-functional-components-nine-wins-you-might-have-overlooked-997b0d933dbc#.u19bbm6em))

blackforestboi:
Hey @AlexMarvelo 

Thanks for getting this running so quickly!
I wonder if you have seen [the mockups I made for the work-flow](https://projects.invisionapp.com/share/KZ8XQZ1BR#/screens/216813851_SETUP_-_Analyse_URLS)
as yours differs so much from it.

The workflow how it is designed in the mockups: 
1. User gets information about download duration & Size
2. User decides if he wants to import bookmarks **and/or** history items and confirms selection
2. Items are requested from chrome history/bookmarks API and stored in the DB (Stage 1 of download, can be expressed in your progress bar, good idea)
3. For each page the HTML is downloaded, parsed and the content stored to the pageObject with that URL. (stage 2 of download)
4. Items where download failed make the counter increase (as seen in the mockups) and flagged accordingly in db. List of failed items is displayed in while downloading runs and afterwards.

In any case, if you have improvements to this work-flow, it is not set in stone. 

Thanks a lot. :)

src/options/layout.jsx
```diff
@@ -1,25 +0,0 @@
...
-        )
-    }
-}
-
-Layout.propTypes = {
-    location: PropTypes.object.isRequired,
-    children: PropTypes.object.isRequired
-}
-
-export default Layout
```
alxmiron, line 25: Layout wasn't changed, just moved to `options/components` folder

src/options/options.css
```diff
@@ -0,0 +1,72 @@
...
+    outline: 0;
+    border: 1px solid #999;
+    box-shadow: 0 0 1px 0 rgba(0, 0, 0, 0.1);
+    transition: background-color .1s ease-in, color .1s ease-in;
+
+    &:hover {
+        background: #cdcdcd;
+        color: #333;
+    }
+}
```
alxmiron: Some styles repeated from [here](https://github.com/aquibm/webmemex-extension/commits/master). I think soon we will combine them

src/overview/actions.js
```diff
@@ -22,7 +22,10 @@ export function init() {
         dispatch(refreshSearch({loadingIndicator:true}))
 
         // Track database changes, to e.g. trigger search result refresh
-        onDatabaseChange(change => dispatch(handlePouchChange({change})))
+        onDatabaseChange((change) => {
+          if (getState().options.searchIndexRebuildingStatus == 'pending') return;
```
alxmiron: Don't handle database changes while search index rebuilding after import - it causes dozens and hundreds of actions that stuck you browser. Maybe we need another solution for this

src/overview/main.jsx
```diff
@@ -8,13 +8,12 @@ import overview from '../overview'
...
-if (process.env.NODE_ENV !== 'production') {
-    ReduxDevTools = require('../dev/redux-devtools').default
-}
+export const ReduxDevTools = process.env.NODE_ENV !== 'production' ?
+    require('../dev/redux-devtools').default :
+    undefined
 
 // Set up the Redux store
-const store = configureStore({ReduxDevTools})
+export const store = configureStore({ReduxDevTools})
```
alxmiron, line 16: We should export `store` and `ReduxDevTools ` to make them reusable on other pages until they are not moved to separate files

alxmiron:
@oliversauter Ok, just haven't seen that mockup.
Will complete it in couple days according to your workflow

obsidianart:
@AlexMarvelo Can you move your suggestion in a more appropriate place? Maybe @oliversauter can help with that. Possibly split them if they can be implemented separately.

blackforestboi:
@obsidianart I'll think of something, good idea.

Alex already mentioned though that he doesn't know where to put it, so I'll take the blame because we have not yet a place for such suggestions :)

Maybe a new issue with the tag code-style would be suitable? @Treora

oliversauter:
Maybe a new issue with the tag code-style would be suitable? @Treora

Treora:
Hi @AlexMarvelo, thanks for the effort and suggestions!

Regarding your questions and suggestions unrelated to this PR, it may be easiest to have a chat (#webmemex on Freenode IRC, or in WorldBrain's Slack if that's easier).

#### Regarding restructuring the options page
You may want to sync with @aquibm, who created the options page, and is also adding a redux store to it. The organisation of components/containers could be changed, making stateless functions of React classes has my preference too, but perhaps this is worth doing as a separate PR.

Note that the options page and overview are currently two completely separate React/Redux apps; this may answer some of your questions about why Redux code and explanation is within the overview folder rather than the root of the source tree. The organisation could still be improved though, as it is indeed rather contrary to people's expectation, and when options page has many parts in common with the overview the situation changes.

#### Regarding making importHistory cancellable
I see you found a solution to make it possible for the user to cancel the history import process, by making the import function a Redux thunk that reads from the state whether to continue. I admire the creativity, but I have the feeling that the feature is maybe not worth the complexity that is created. Importing the browser history (not talking about fetching the pages here) should in my experience take at most a few seconds, and for an MVP I deem it acceptable to not have a cancellation option for that. What do you think?

If we do want to cancel the process (now or later), it may be worth looking into good programming patterns to achieve this, preferably without making the import function too coupled to the rest of the application (to keep it reusable and maintainable). At the least, the checks could be factored out of the import function by just passing it a second argument `shouldContinue: () => (getState().import=='pending')` (or similar). Looking a bit further, redux-saga [has a cancellation pattern](https://redux-saga.github.io/redux-saga/docs/advanced/TaskCancellation.html) built in, but I'm not too eager to adopt sagas just for that. I am not too enthusiastic about creating [a solution with redux-observable](https://redux-observable.js.org/docs/recipes/Cancellation.html) either, as it does not seem to make code more readable. Doing some simpler generator magic might work; I am not an expert at this.

By the way, in any case, an improvement would be to convert the promises to use `async`/`await` syntax (#7).

As for the performance of the import process, I wonder if it helps if the browser api is called in sequence rather than in parallel when searching the visits to each historyItem.

alxmiron:
Hi @Treora and @oliversauter 

PR description was changed. Check it, please

**Regarding making importHistory cancellable**
I agree that promise cancellation need a better solution.
I can suggest [bluebird](http://bluebirdjs.com/docs/api/cancellation.html) as it's easy to integrate in our project and has obvious behavior, unlike redux-observable. Anyway, such work should be done in separate issue.

I want to underline, that this PR is about UI, all data is fetched from props only. Solution for hard drive size and time estimating, loaded pages amount etc require weighty improvement in import chain.

Treora:
@AlexMarvelo: Hi, this PR was going in the right direction, did you by any chance work on this further after our last chat? Or planning to? I suppose some things may be easier to tackle after (or in harmony with) the blacklist feature (#20, #84) , since it also adds a redux store to the options page.

poltak:
I've continued on from this in the [`feature/import-ui` branch](https://github.com/WorldBrain/WebMemex/tree/feature/import-ui/) in the WorldBrain fork. Done a bit of refactoring so far, but will be able to use the presentational and state code from @AlexMarvelo's commit

Treora:
Seems like this PR can be closed then.

PR #67, opened 2017-03-13
Delete page entries from db for single url.

wimpykid26:
@Treora What I did was:
Used findpagesbyurl and normalizeresults from deduplication branch.
get the corresponding entries
called db.remove

Please review. If there is any error I'll fix them ASAP.


wimpykid26:
@Treora any thoughts?

Treora:
It looks like this PR also contains the bookmarks import code and other stuff that is not related to this issue, so the diff is too messy to be reviewed.
By the way, I merged deduplication into master, so if you rebase onto master you can also use the `findPagesByUrl`.

wimpykid26:
@Treora Apologize for the mess. Something went wrong while creating the branch. cleared the diff. used `findPagesByUrl` to get corresponding entries of same url.

src/delete-pages/delete-pages.js
```diff
@@ -0,0 +1,29 @@
...
+      return updatePageSearchIndex()
+    }).then(() => {
+      console.timeEnd('rebuild search index')
+    }).catch(err => console.log(err));
+  }
+}
+
+export default function deletePage(query) {
+  console.log(query)
+  var result = findPagesByUrl({query, limit: defaultResultLimit}).then(
```
Treora: `findPagesByUrl` does not take a `query` or `limit` argument.. how has this worked for you at all?

src/delete-pages/delete-pages.js
```diff
@@ -0,0 +1,29 @@
+import db from '../pouchdb'
+import { findPagesByUrl, updatePageSearchIndex } from '../search/find-pages'
+
+const defaultResultLimit = 30
```
Treora: Why a result limit?

src/delete-pages/delete-pages.js
```diff
@@ -0,0 +1,29 @@
...
+
+//Delete all Page entries returned by findPagesByUrl
+function deletePageEntries(pagesResult) {
+  console.log(pagesResult);
+  const pages = pagesResult.rows.map(row => ({
+    id: row.doc._id,
+    rev: row.doc_rev
+  }));
+  for (var key in pages) {
+    db.get(pages[key].id).then(doc => db.remove(doc._id, doc._rev)).then(() => {
```
Treora: Why fiddle around with `id`s and `rev`s and get each page again? Just running `pagesResult.rows.map(row => db.remove(row.doc))` should work, no?

src/delete-pages/delete-pages.js
```diff
@@ -0,0 +1,29 @@
...
+      console.timeEnd('Deleting pages end')
+      console.time('rebuild search index')
+      return updatePageSearchIndex()
```
Treora: You don't want to reindex after every single deletion. Perhaps use `Promise.all` on the return value of the pages.
Or, equivalently, replace the promises with an `async function` that `await`s each removal.

Treora (commented):
That is a lot easier to oversee indeed. The code looks rather broken though; has it even worked for you? Please first get things to work before asking for review.

wimpykid26:
screenshot from 2017-03-21 01-39-12
@Treora Cleaned the code as you said. Made the pagesResult return value into a promise and rebuilded search index after the promise is returned.

Treora:
The essence is not bad, but it's still only half the solution here; closing due to inactivity.

PR #68, opened 2017-03-17
Avoid cross-folder relative imports

Treora:
Use the `src` folder as a phony npm module to avoid `../../..` imports.
Relative imports within a subdirectory of `src` are left as is.

There seem to be [many hacky approaches](https://gist.github.com/branneman/8048520) to work around this ugly issue. I'm following [slorber's approach](http://stackoverflow.com/questions/20158401/how-do-i-manage-relative-path-aliasing-in-multiple-grunt-browserify-bundles/23608416#23608416) of using browserify's `paths` option. I decided to keep the `src` prefix to each folder in order to make clear they are not external modules.
Thanks @AlexMarvelo for the suggestion.

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -40,6 +40,7 @@ const files = [
 const browserifySettings = {
     debug: true,
     extensions: ['.jsx', '.css'],
+    paths: ['.'],
 }
 
 function createBundle({entries, output, destination, cssOutput},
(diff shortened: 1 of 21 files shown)


obsidianart:
I really like this technique as it makes refactor easier (you can move files around quicker).
On the other side, when I implemented, I had 2 people complaining because this breaks webstorm path resolving (which means cmd+click or "go to source" doesn't work well, and other advantages in debugging and getting support from the IDE might get lost).
I'm not against this technique, neither I consider it an "hack" since that's what you normally do with the node modules.
Those are the pros and cons I'm aware of, feel free to merge if you want to, it's easy to revert in the future if it creates more issues than advantages.

Treora:
Thanks for the feedback, always good to hear someone's opinion before merging it.

In the IDE problem you describe, if the IDE does look in the `node_modules` folder, one could perhaps create a symlink: `ln -s src node_modules/src`

obsidianart:
the IDE looks in the node_modules folder. what the ide can't do is knowing that the "src/" is mapped to a specific folder, since this is browserify configuration.

Issue #64, opened 2017-03-12
Represent search query in the url of overview


blackforestboi:
Right now, when opening the search overview, it has no url whatsoever. Also when typing in a search query its not represented in the URL.

I think it'll be helpful to represent the search in a url, as it is a common work flow and also would help to link to the search results from the address bar.

Rohanhacker:
I don't think it's possible

Rohanhacker:
the absolute path of overview is something like this chrome-extension://<extensionID>/overview.html
but I am not sure if one can represent it.

Treora:
Good to put this issue on the list. Helps indeed with searching in the address bar (#14), and to not lose the result list when using the back button (#52).

@Rohanhacker: We can indeed not change the paths, but we can however add a query parameter or fragment identifier. Something like `overview.html?q=bananas` would be fine.

Not sure whether to use the react/redux router for this. Possibly. What we need is bidirectional syncing between a field in the redux state and a query parameter in the URL.

Rohanhacker:
I'll give it a shot with react router

Chaitya62:
Can we use window.location.hash? or something similar?

Chaitya62:
I' ll work on this

Chaitya62:
#63 already address this I guess

blackforestboi:
@Chaitya62 No #63 is only to link up the search iteself with the address bar.
What is needed here is that the search query is represented as a URL, when searching in the overview. 

something like `chrome-extension://adfasfsadfsadlfakfdjasdfasdf?search="test"&from="03-03-17"`

Chaitya62:
@oliver So should I work on it also does using ```window. location.hash``` sounds good using that we can have something like ```thecurrenturl#search=searchQuery``` or something similar

Treora:
@Chaitya62: I would rather not have to manually read and write the hash, but rather use some clean abstraction to sync a designated part of the application state with it.

> Not sure whether to use the react/redux router for this. Possibly. What we need is bidirectional syncing between a field in the redux state and a query parameter in the URL.

blackforestboi:
@Chaitya62 It seems as there is still some work to do with your old PRs? Right?
Taking the work on the Research-Engine repo into account would be the 4th mini-project you work on simultaneously :) 

For simplicity reasons, I would suggest first completely finishing your old PRs.

Chaitya62:
[...]

PR #69, opened 2017-03-19
Represent search query in the url of overview #64

Rohanhacker:


Treora:
I finally got around to look at this PR. Nice work, it seems to work as intended. I did however wonder about the maintainability of the code, since care has to be taken that both the window location and the Redux state keep the same value. The logic for this has been put into the overview component, while conceptually it has very little reason to be there: it is a problem of syncing redux state and window location. Your approach matches a more commonly suggested approach (e.g. [here](http://stackoverflow.com/questions/36596996/how-to-sync-redux-state-and-url-query-params/36657751#36657751)), but starting from this idea I decided to try to decouple this logic from the UI, and make an abstraction that will also allow us to easily expose other variables from the state to the location. See PR #92 for the result. Closing this one. Thanks nevertheless!

Issue #45, opened 2017-03-02
Log time spent in each visit


Treora:
We currently only log the start time of each visit, not the end. This would be an obvious first thing to add. Much better still, we should log when a tab becomes active or inactive, because that is what actually matters, and possibly what we should really call a visit.

Let's start with logging the end time of the visit (= when the tab closed), and counting the total time spent viewing the page (≈ tab being active) and adding this number as a separate field to the visit.

This may require keeping some state in the background script about which tabs correspond to which visits.

If you have an idea or existing code for approaching this, feel free to share.

Treora:
Bonus points for adding this idea from #11: use mouse or keyboard activity to estimate whether the user is looking at the page or is away from keyboard, as to not log the latter as active viewing time.

anona-R:
Can we use [ifvisible]( https://github.com/serkanyersen/ifvisible.js)  to check whether a user is looking at a page or not and then track the time by a timer

Treora:
> Can we use ifvisible

Yes, brilliant, exactly the piece we need for the bonus points. What remains is just counting and summing up active times in the [activity-logger content script](https://github.com/WebMemex/webmemex-extension/blob/master/src/activity-logger/content_script/index.js), then somehow report it to the background script (not sure yet which way is best for that, but that can come later).

vaibsharma:
I made a chrome extension two months back, in which I required the total time spend in the current tab or website in that tab. We can use that approach for letting the end time come into the picture . By using port.postmessage({}) API of chrome we can send data whenever the condition satisfy . And then listen to this message in background .
 ^but I am not sure because in that I knew the websites that were allowed by the content script and each one of them had their indices , but since here we are using content script for all then perhaps we have to think of using different idea for using that code

vaibsharma:
^but I am not sure because in that I knew the webaites that were allowed by the content script and each one of them had their indecies , but since here we are using content script for perhaps we have to think of using different idea for using that code

NimmiW:
@Treora @vaibsharma yes we have think of some different idea. Let me and @uchithaR look into this :) We like to contribute to this project.

NimmiW:
@Treora https://github.com/NimmiW/activity-log-time/blob/master/background.js

Here we @uchithaR have done getting the active time log of all the tabs when navigating through tabs. This uses that active status of the tab to get the active time. Now we have work on detecting the real active time by using mouse over events and keyboard pressed events. 

[...]

PR #70, opened 2017-03-19
[Work-In-Progress] [Do Not Merge] Log time spent in each visit

NimmiW:
This PR is regarding the issue at https://github.com/WebMemex/webmemex-extension/issues/45


NimmiW:
Sample output: (background script)

capture

This reports the time spent on tabs opened.

I need a guide to make them save permanently, so that later on we can use those details to do analytics etc :)

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
     maybeLogPageVisit({url, tabId})
 
 })
+
+
+//********************************************************************************/
+var timestamp = new Date().getTime;
```
obsidianart: this is probably a const, but it seems unused

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
...
+
+/*chrome.runtime.onMessage.addListener(function(response,sender,sendResponse){
+    alert(response);
+});*/
+
+
+browser.tabs.onCreated.addListener(function (tab){
+    number_of_opened_tabs = number_of_opened_tabs + 1;
+    //console.log('total opened tabs: ' + number_of_opened_tabs);
+    var t = {
```
obsidianart: this would be a const, but t isn't a great name. You can find a good name or push directly the object.

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
...
+        id : tab.id,
+        activetime : 0,
+        isActive : false,
+        lastActivatedtime : Number(new Date().getTime())
```
obsidianart: Date.now() is probably easier to read. Why using type coercion to Number?

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
...
+    };
+    tabList.push(t);
+    console.log('a new Tab created : ' + tab.id);
+})
+
+browser.tabs.onRemoved.addListener(function (id){
+    number_of_opened_tabs = number_of_opened_tabs - 1;
+    //console.log('total opened tabs: ' + number_of_opened_tabs);
+    for (var index in tabList){
+        if(tabList[index].id == id){
```
obsidianart, line 44: do you expect more than one match? if not, you should probably use array.find instead.
Also, is there any reason to use an array instead of a hashmap using the tab.id as key?

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
...
+    console.log('a Tab was closed : ' + id);
+})
+
+browser.tabs.onActivated.addListener(function (tab){
+    //console.log('Tab activated : ')
+    //console.log(tab);
+    for (var index in tabList){
+        if(tabList[index].isActive == true){
+            tabList[index].isActive = false;
+            tabList[index].lastActivatedtime = Number(new Date().getTime());
```
obsidianart: as above about Number etc

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
...
+            console.log('a tab went inactive : ' + tabList[index].id);
+        }
+        
+    }
+    for (var index in tabList){
+        if(tabList[index].id == tab.tabId){
+            var currTimestamp = Number(new Date().getTime())
+            var difference = currTimestamp - tabList[index].lastActivatedtime;
+            tabList[index].activetime = tabList[index].activetime + difference;
```
obsidianart: ```tabList[index].activetime += difference;```

src/activity-logger/background/index.js
```diff
@@ -9,3 +9,68 @@ browser.webNavigation.onCommitted.addListener(({url, tabId, frameId}) => {
...
+    //console.log(tab);
+    for (var index in tabList){
+        if(tabList[index].isActive == true){
+            tabList[index].isActive = false;
+            tabList[index].lastActivatedtime = Number(new Date().getTime());
+            console.log('a tab went inactive : ' + tabList[index].id);
+        }
+        
+    }
+    for (var index in tabList){
```
obsidianart, line 62: this is the second loop and there's no reason for it, you can have 2 different behaviour in the first loop
```
if(tabList[index].id == tab.tabId){..}
else {...}
```
That said, I would still evaluate hashmap instead (it is not necessarily better)

NimmiW:
Thank you @obsidianart I will make those things you showed. :)

src/activity-logger/background/index.js
NimmiW: I used `new Date().getTime()` so that we can easily get the number of milliseconds that a particular tab is activated. Is it a good choice?

src/activity-logger/background/index.js
NimmiW: I want a single match. I think it is better use a hashmap. I will correct that :)

src/activity-logger/background/index.js
obsidianart: They added the static method for simplicity, they are equivalent.
Type coercion isn't needed, they are already numbers
```new Date().getTime() === Date.now()```

File: src/activity-logger/background/index.js (hunk 4 above)
Comment (obsidianart) [reply src/activity-logger/background/index.js:44]: true, but you also want to loop to deactivate it, so do as you think it's easier to read.

**Comment by obsidianart** [comment]
@NimmiW Nothing major but since you are here for learning more ways of doing things it's good to be picky.

File: src/activity-logger/background/index.js (hunk 3 above)
Comment (NimmiW) [reply src/activity-logger/background/index.js]: yes I put Date.now(), it also worked nicely :) @obsidianart thanks for the tip

File: src/activity-logger/background/index.js (hunk 7 above)
Comment (NimmiW) [reply src/activity-logger/background/index.js:62]: I want second loop work to be run after the first loop work. 
is there anything I can do for that? I think probably including a hashmap might reduce the complexity

Comment (obsidianart) [reply src/activity-logger/background/index.js:62]: you don't need 2 loops, you can put the if inside the first loop.
If you have an hashmap you avoid the second loop anyway, and the first will be
```
Object.keys('tabHash').forEach(key=>{
    let tab = tabHash[key]
    //do what you currently do
})

tabHash[tab.tabId] = //second "loop"
```

in contrast with the current (written using array methods)
```
tabList.forEach(_tab=>{
    if(_tab.id == tab.tabId){ //this is what you do in the second loop}
    else {//this is what you do in the first loop}
})
```

It is worth it studying array methods regardless of the solution: forEach, map, filter

NimmiW:
Ok @obsidianart I Will go through my code again and correct those things you mentioned. :)

Treora:
Closing due to inactivity. Feel free to reopen (or make new PR) if you do get this going. Until some proper basis for keeping state (#91) is there, it may however be easier to just consider this work exploration of possibilities. Which is also useful!

PR #74, opened 2017-03-23
Added date time to filter search.(Continued from PR #66 ) for issue #26

RajPratim21:
@Treora  @oliversauter 
because of certain version control reasons previous version was behaving strangely now I have added all the stuff in this repo to keep the code clean and understandable, for issue  #26

PR #75, opened 2017-03-24
Natural language date parser added.

RajPratim21:
@oliversauter  @Treora , I have added the Natural language date parser , using moment.js
naturalparser

PR #76, opened 2017-03-24
Date picker for time range selection

### LINKED ISSUE #26: Filter search results by time
Oliver and I discussed two ways to express the desired time range:
1. by text query: one could write "since March", or "last week" (like WorldBrain&Falcon already do it)
2. by a separate input; perhaps a [nice range slider](https://refreshless.com/nouislider/) or a [date picker](http://www.daterangepicker.com/)

Both are nice, but we can start with just one.

Treora:
This builds further on @RajPratim21's work of #74, to resolve #26. Main changes:
- Make search independent of the store again (pass date range as arguments).
- Clean up code and css.
- Style UI a bit.
- Refresh search results when date range changes.
- Also apply the date filter when no search query is given.

Comments are welcome, otherwise I will merge it as is in a few days.


RajPratim21:
@Treora , Thank you for your help, Once you merge it I will make the required changes in the PR #75  for natural language date parser, as they are working in same style, to maintain the integrity.

Issue #28, opened 2017-02-14
Batch fetch&analyse pages


Treora:
Download a page to analyse its content and update its entry in the database.

Intended for getting more knowledge about imported browser bookmarks&history. See WorldBrain's corresponding [issue](https://github.com/WorldBrain/Research-Engine/issues/92).

bohrium272:
Granted we have to obtain page content using `XMLHttpRequest`, there would be a whole lot of requests. I was thinking if there is any library/module to do it with optimal performance. 
Found these:
* https://github.com/volpav/batchjs
* https://github.com/socialradar/batch-request

If we can work something out here, we should use it for WorldBrain too.

Treora:
There will be a whole lot of requests indeed; but each to different servers, and servers that are not ours, so unfortunately those libraries seem of no use in our case.

What I am thinking of is to make a function `fetchAndAnalyse` that, given a url, will end up putting a `page` object in the database that is as similar as possible to what would have been produced if the user would just have visited that page manually.

Having such a function, we could then apply it to every url in any given list of urls, in particular the urls imported from one's browser bookmarks and browser history. This would be the 'batch' part, and it may take a while for larger lists, so the main tasks here are to make it possible to pause and resume the process, and to inspect its progress.

But the first task is not the batch aspect, but to fetch&analyse a single page. We could start with just storing the html (for html documents), extracting its plain text for indexing. Later we can also run analysis on it to extract metadata, main text content, ~~maybe make a screenshot using [PhantomJS](http://phantomjs.org/)~~ or such.

Now I am thinking of it, I may be making it too complex; if we are writing this to run in the browser anyway, we could skip the PhantomJS, and avoid having to make a separate analysis pipeline for these pages that are fetched in the background. We could simply open the url in a tab programmatically, have it be analysed as usual, close the tab programatically, and then continue with the next one. It also makes the process very visible to the user, which may be a good thing, though there may also be complications with this approach.

We could start with implementing whatever seems easiest to implement, and experiment a bit to get a better feeling for it.

blackforestboi:
> extracting its plain text for indexing.

In the WorldBrain code I used a code snipped that I extracted from a library called textract:
https://github.com/dbashford/textract

The cool thing is that the library is capable of reading all sorts of formats as well. I just used the HTML part. 

HTML, HTM
Markdown
XML, XSL
PDF
DOC, DOCX
ODT, OTT (experimental, feedback needed!)
RTF
[...]

PR #79, opened 2017-03-29
PDF support added

the-fallen:
As mentioned in issue #27 ,
This includes module for pdf extraction and some minor changes in page analysis to support pdfs, 
For testing,
Do npm install to get pdfjs-dist library(changes in package.json), or run 
`npm install pdfjs-dist --save`
After make, to see it work, see background.js output in console to see text and metadata objects.

As @Treora mentioned about the two situations about user visiting a document and just having the url,
In case the file is in cache, XMLHttpRequest gets file from cache without downloading again.
Also, pdf worker added in extension folder doesn't look clean, i know. But PDFJS library uses document.currentscript to load worker script which is disallowed in content scripts. So this is a workaround i had to do to get it working, i'd be glad if someone could suggest a cleaner fix.


src/page-analysis/content_script/extract-page-metadata.js
```diff
@@ -1,10 +1,18 @@
 import { getMetadata, metadataRules } from 'page-metadata-parser'
+import extractPdfMetaData from 'src/pdf-analysis/extract-pdf-metadata'
 
 // Extract info from all sorts of meta tags (og, twitter, etc.)
 export default function extractPageMetadata({
     doc = document,
     url = window.location.href,
 }={}) {
+
+    if(url.substr(url.lastIndexOf('.')+1) === "pdf" ){
```
oliversauter: maybe the substring `.pdf` makes more sense, as we can be sure that URLs that mention "PDF" are note wrongly redirected.

blackforestboi:
I tested it the text and metadata extraction works! (although not searchable due to the current bug of the search not working)

src/page-analysis/content_script/extract-page-metadata.js
Treora, line 10: Tip for next time: `if (url.endsWith('.pdf')) {`

This may be a good start, but I'd like to have a better test than just checking the url. The check may also have to happen elsewhere in the program flow.

src/pdf-analysis/extract-pdf-text.js
```diff
@@ -0,0 +1,53 @@
...
+function getText(blob) {
+    return new Promise(function (resolve,reject) {
+        var fileReader = new FileReader();
+        fileReader.onload = function (blob) {
+            require('pdfjs-dist');
+            require('fs');
+
+            // workerSrc needs to be specified, PDFJS library uses
+            // Document.currentScript which is disallowed by content scripts
+            PDFJS.workerSrc = chrome.extension.getURL("pdf-worker/pdf.worker.js");
```
Treora: We use the `browser` API, not `chrome`, for cross-browser compatibility. Use the [MDN docs](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/) as reference. Also I think those docs recommend to use absolute paths.

Treora (commented):
Nice, this looks like a good start. Some thoughts:

I like how you make only as few changes as possible to the existing analysis code, but I think it may be a big waste that the pdf is now loaded twice, once to extract the metadata, once to extract the text. I was assuming that pdf analysis would have to be a separate flow altogether (that would be run instead of the page-analysis). As you show however, it could share the same flow, and just handle the individual functions differently. If we keep that approach, we should change the flow a bit though. We would want to add a cache of some sorts for the parsed pdf object to not load it twice, and perhaps also do the check to see whether it is a pdf only once (this could be a more generic test for the document type). I'm not sure which approach is best.

One consideration here is still if/how we handle pdfs that are not opened in a tab, but perhaps in an external application. We could still analyse them, and would ideally reuse most of the code, but in that case we may want to not use the page-analysis flow; or we should rethink its approach so it also works without giving it a tab. The flow here would be similar to doing a fetch&analyse in the background (#28). I don't mind if we would at first only support pdfs opened inside a tab, but it's good to think ahead.

Regarding the PDF worker. I understand the need for a workaround to point at the worker script. We should be able to do this without copying the code from pdf.js into our repo though. An option is to have `extension/pdf-worker/pdf.worker.js` be built from a separate source file that simply performs `import 'pdfjs-dist/build/pdf.worker.js'`. (or `.min.js`).

src/page-analysis/content_script/extract-page-metadata.js
the-fallen: I believe the only way to know for sure it's a valid pdf document is to have a check when PDFJS.getdocument returns.

the-fallen:
Thanks for the browser API correction(updated that) and pdf worker suggestion!

Also, to solve both problems at once, double pdf loads and double pdf checks, maybe we can have a single `extract-pdf-data.js` inside `page-analysis/content_script` which returns a promise for an object containing metadata and text both, and have a single pdf check in `page-analysis/background/index.js` then decide whether to use `extract-pdf-data.js` to set both text and metadata at once in case of pdf, or use `extract-page-text.js` and `extract-page-metadata.js` in case of normal page as usual.

And I'll be working on my project proposal for a few days, so will implement this asap after getting done with that.

Issue #71, opened 2017-03-19
Suggestion: Travis CI integration


NimmiW:
This is a suggestion. 

If we can implement a continuous integration for webmemex, I think it would be beneficial for the developers. 

Any comments or thoughts about that?

Since I am new to open source, your thoughts help me to improve me :)

obsidianart:
it would be useful, especially if you would be able to do the github integration and run test before merge :)

obsidianart:
check with @oliversauter , there's already a plan for CI

blackforestboi:
@danrl is taking lead on mentoring this as a GSoC project. 
Here is the Gitbook we use to document our testing efforts: https://www.gitbook.com/book/worldbrain/project-ideas-continous-integration-pipeline/details

danrl:
TravisCI integration sounds nice, but what tests do you like to run? It's the lack of tests that we have to tackle first.

obsidianart:
Even one test would be good, or just compile it. They are 2 problems and it's good to address both.

Treora:
I have little experience with CI, and I may yet have to understand the current relevance of it for this project, but I would be happy to help set things up if any of you want to take the lead.

My first question is: what exactly would we want to achieve here? The ability to check if your PR breaks things when merged? Making the health of PRs/branches visible to everyone?

My thoughts on this, feel free to correct me:

- Without any checks to run CI seems a bit precipitous.
- Before having written tests, a good first check to run would be a linter (#72).
- We should first focus on running checks locally; e.g. add a `make test`. Quick checks, or at least the linter, could perhaps be integrated into `make watch`.
- When locally running checks works, we could employ/deploy a service that runs the checks automatically on every branch and PR.
- I would prefer to employ a service that could be self-hosted if desired, rather than buying into a single provider. Some questions: Are there more Travis-compatible services? Is it feasible to self-host it? (some answers may be [here](https://github.com/travis-ci/travis-ci/issues/2114) or [here](stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally)) Would [Jenkins](https://jenkins.io/)/[Jervis](http://sam.gleske.net/jervis/) be friendlier in this regard? What else is there?

reficul31:
>My first question is: what exactly would we want to achieve here?

[...]

Issue #72, opened 2017-03-20
Add a linter


Treora:
To automate code quality checking.

- [x] Choose a linting tool (JSHint, ESLint, ...?).
- [x] Configure it to what seems a good coding style; preferrably endorsing some existing standard.
- [ ] Run as a check in `make test` (yet to be created).
- [x] Run and print warnings in `make build`.
- [x] Run and print warnings in `make watch`.

obsidianart:
I would try to have a linter which autofix things (like tabs and spaces) when you commit

reficul31:
I would like to take up this issue as I have already made some headway after having a brief chat with @Treora . According to me the steps to follow are.
1. Observe and document the coding style
2. Make a ESLint config file with the coding styles observed
3. Bootstrap the linter and run tests for all the packages already made.

I intend to do this as an experiment to get to know the coding style a bit better and so that it also helps in the CI we are trying to set up.

reficul31:
Following my experiment I have tackled some problems. As per the suggestion by @Treora I found a standard we should use for the project, but right now the code is not at all consistent. As I see it I can go two ways. Either fix the code to suite the style or fix our own custom standard which again has some problems as the code isn't consistent throughout.
Any feedback will be extremely helpful. To follow the [progress](https://github.com/reficul31/webmemex-extension/tree/issue_72)

reficul31:
So I have made some significant progress. I have already added the eslint extending the standard config provided by default. I have also integrated the eslint-react plugin to allow jsx to be parsed. To see the [progress](https://github.com/reficul31/webmemex-extension/tree/issue_72). If I could get the go signal to refactor the styling I will start off by changing the style of the src module first. Thanks.

Treora:
@reficul31: Good progress indeed! I would not mind changing some code style by the way, e.g. using two-space indentation is fine with me if those standards all do that too.

I just did quite a refactor to use the async/await syntax throughout the code. Shall we first merge that and then fix the styles?

reficul31:
@Treora I just now added a bit of cleaned code. Please review it once you are free and see if the style suits the project.
>  I would not mind changing some code style by the way, e.g. using two-space indentation is fine with me if those standards all do that too.

It has been changed in the most recent commit. Also it doesn't matter as soon as we put an editor config in place the indentation error will be sorted out itself.

>I just did quite a refactor to use the async/await syntax throughout the code. Shall we first merge that and then fix the styles?

[...]

PR #80, opened 2017-03-31
Added Travis CLI for build Automation ( Ready for merge )

niroshn:
#71 
Successfully added Travis build file ( .travis.yml )
It is working fine.

After every PR travis automatically run these build.
Then Developer can catch their conflict and error.


niroshn:
When Running Build tassk
travis-build

niroshn:
After successful build 

travis

.travis.yml
```diff
@@ -0,0 +1,22 @@
+sudo: false
+
+language: node_js
+
+node_js:
+  - "4"
```
obsidianart, line 6: why 4?

obsidianart:
@Treora is "make" the right script? can we change it now to npm run make if that's the case so the automation can rely on npm scripts?

Treora:
> @Treora is "make" the right script?

My tradition is to put a Makefile in every single project, so that users don't have to be bothered which command to run (like a [universal install script](https://www.xkcd.com/1654/)). Also it serves as a reference of the possible actions one may want to do.

For this configuration file, calling the `npm` or `gulp` commands directly would be fine too. Calling `make` does however have the advantage that the Makefile is also tested.

.travis.yml
niroshn: This is version of node.js

niroshn:
@Treora Is there any other issue in this PR

Treora:
> @Treora Is there any other issue in this PR

#71

Treora:
Oh you mean literal issues. :)

I don't know anything about Travis config files, so I cannot judge. We had not really dediced whether to use Travis, or another platform, so that is the main question still.

niroshn:
yes literal issues.

Travis is a most popular CI platform. 
After trying Travis CLI we can remove it If you need.

niroshn:
I added eslint in here #72

danrl:
It's not doing any harm using TravisCI for a while. We can switch whenever we like.
I support the move to use TravisCI and evaluate after a while if it was helpful and improved code quality or not.

.travis.yml
obsidianart: yes, why 4 and not 7?

obsidianart:
@Treora I would say ```npm run``` and ```npm build``` are more standard in js projects than make

Treora:
It would indeed not be a problem to start with Travis. I just thought now would be a good moment to look at the options, though I have not deemed CI important enough to investigate myself.

As for `make` versus `npm run` and such, `npm` is indeed more standard in javascript projects nowadays; however there is more than javascript, and more than nowodays. Anyway, I have no preference for whether this script runs `make` or `npm run`.

I would remove the eslint from this PR, that's being worked on in @reficul31's [branch](https://github.com/reficul31/webmemex-extension/tree/issue_72). Also is there a reason for the `npm-debug.log` to be there?

niroshn:
@Treora  Shall we try with Travis CL then we can decide continue or not

.travis.yml
niroshn: Yes We can use 7 . 
 4 mean latest 4.x release
7 mean latest 7.x release
Will update PR for 7

bwbroersma:
@Treora also look into [drone.io](https://github.com/drone/drone), it's very easy to self host.

Treora:
@bwbroersma: thanks for the Drone tip. It looks like it requires a couple of extra steps though, so I opted for trying Travis for now, as I am not even convinced yet of the added value of these build&test services.

@niroshannrsh: as eslint has been added in #83, I just looked at your first commit in this PR. I only saw the need for two of the lines in your `travis.yml` config, so I started from scratch instead. I wonder what was the reason to add the config for `addons.apt`, and for the `install` commands?

Issue solved in f317827, closing this PR.

PR #83, opened 2017-04-03
Issue #72 -- Add a Linter

reficul31:
#### Features of the PR.
* Adds eslint to the project for consistent styling
* Adds editor config to the project
* Configures to remove most style errors and warnings.

@Treora Please take into account that the PR still has some errors. Please either remove the errors or remove the lint itself. Please check and see if the rules are to the maintainers liking, or they can always be changed. Thanks.

#### USAGE
* sudo npm run lint -- for checking the errors in the lint
* sudo npm run lint-fix for automatically fixing the errors


.editorconfig
```diff
@@ -0,0 +1,33 @@
...
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+# Styling applicable to the file extension given
+[*.js, *.jsx]
+# This is due to the fact that github allows only 119 columns of space to view
+max_line_length = 119
```
obsidianart: I would really discuss this one

reficul31: Actually this is a widely observed standard in many open source projects
```
max_line_length = 119
```
This is done so that there is no need for horizontal scrolling while reading the code as github provides only 119 columns worth of space to read the code. But I can change if it seems too restricting.

obsidianart: @Treora makes the final decision, I wouldn't have this rule

bwbroersma: The reason here is actually 120, but minus 1 char for the diff view. Why 120 I don't get because it seems pretty much browser specific, and people just choose a round number, see:
screenshot of github editor in different browsers on OsX
(source [Django developers discussion](https://groups.google.com/forum/#!msg/django-developers/3o7i18ewMWs/JaCaf7Cjef0J))
For more random numbers, see [this stackoverflow thread](http://stackoverflow.com/q/22207920).

I do agree setting line lengths to something higher than 80, but 119 and binding that to Github web view and some browser specifics and then rounding that number seems pretty ... arbitrary.

obsidianart: I understand github diff works on this length, I'm not discussing the number, I'm dicussing having any number there.
If tomorrow (real life story) I need to put an svg encoded icon somewhere I do want it on one line and I don't care if you have to scroll.
I'm not really worried of a dev writing a very very long line, why would it do that? So I consider this rule more a possible issue than a safeguard.

reficul31: Point noted, I shall remove this line immediately. Or would you like another value to takes its place?

reficul31:
There are 10(5 warnings and 5 errors) persisting errors that I cannot resolve as they require some big design decisions. It would be really helpful if someone could help me with them.

##### no-unused-vars error
This error finds its [source](http://eslint.org/docs/rules/no-unused-vars). It is currently set to warning. Places of occurrence
> src/activity-logger/background/log-page-visit.js => (finalPagePromise , visit) both are unused vars

##### react/no-unused-prop-types error
This error finds it [source](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unused-prop-types.md). It is currently set to warning. Places of occurrence.
> src/options/layout.jsx => (location, children) both are unused prop types

##### prefer-promise-reject-errors error
This error finds its [source](http://eslint.org/docs/rules/prefer-promise-reject-errors). It is currently set to error. Places of occurrence
> src/util/event-to-promise.js => line 58

##### promise/catch-or-return error
This error finds its [source](https://github.com/xjamundx/eslint-plugin-promise/issues/22). It is currently set to error. Places of occurrence
> src/util/tab-events.js => line 29

##### handle-callback-err error
This error finds its [source](http://eslint.org/docs/rules/handle-callback-err). It is currently set to error. Places of occurrence
> src/util/webextensionRPC.js => line 27 and 17
> src/util/when-all-settled.js => line 3

These are all the errors that are still there. If we could remove all the errors or remove their lint rules then the code will be fully clean. I will continue looking for more rules that can help make the project more consistent. Thanks. Once the rules are finalized and integrated I will continue to add the git hook.

.editorconfig
Treora: I intended to ask about it too; adjusting rules to Github's presentation seems somewhat arbitrary (or submissive ;) ) indeed.

I have myself tried to keep things at a soft limit of 80 characters, though I'd be fine with a larger number too. As said it is nice to be able to make longer lines if needed; I am not sure whether editors interpret such a rule as soft or hard.

Treora:
Thanks for the PR. I will do another round of changes to remove (or suppress) the remaining warnings/errors.

.editorconfig
obsidianart: If we have a linter it will probably stops you from committing if you have linter errors, so I would say it will be pretty hard :)

Treora:
Done & merged. GitHub seems not to have detected that, so closing this manually. Thanks for the effort!

reficul31:
@Treora thanks for all the help!!

Issue #52, opened 2017-03-05
When clicking on an entry, open a new tab, instead of the same one


rusuo:
When clicking on an entry from your history, currently it is opened in the same tab.

I would suggest to be opened in a new tab, for having an easier to come back to the history, or to check what is happening

blackforestboi:
Good idea! 

Actually right now I can't open the results in a new tab with cmd+click. I have to deliberately right click and then on "open in new tab", which is quite cumbersome, especially when you want to open more than one result.

Treora:
I personally dislike making the same-or-new-tab choice for people, and every history view or search results page I know does not open results in a new tab by default. The items are just links, so Ctrl or middle-mouse click should just open them in a new tab as usual. I wonder why it does not work for you with cmd+click.

More generally, I am still thinking how and when to present the overview anyway. In every new tab is a choice that may be reconsidered, and also only works in Chrome|ium. Seeing your history when closing a tab may be useful, or having it as a permanent special tab (like a 'pinned' tab). Decisions on this can be made and changed later, user research is welcome.

anona-R:
I will open a PR soon on this 👍

blackforestboi:
@Treora 

I just tested it also on Firefox and have the same issues. Might be ok to make the choice for now and let people make a decision later on.
With the PR @uchithaR just made it also not working, so it does not really solve the issue of being able to open multiple ones quickly. 

MIght not be the most common use case after all. Still, i would leave the open-in-new tab for the reason that if people make a search, they would have to go back and retype everytime they visited the wrong page and redo the search.

Treora:
@oliversauter: Hah, I know what causes the issue for you: Cmd key is mapped to the Meta key, which I [used for debugging](https://github.com/WebMemex/webmemex-extension/blob/4ab650850cecdc66b94eb64cb21da798815cef48/src/overview/components/VisitAsListItem.js#L20). Will change this.

Treora:
Doing the debugging on ctrl+meta now, as it seemed quite a handy feature to keep. I hope that is not needed for something..? (opening in a new window is cmd+shift on Mac, I suppose?)

blackforestboi:
Sometimes its the simple things :)

> I hope that is not needed for something..?

on mac it does a right click and opens the context menu. Seems less problematic, didn't even know I can open the right click menu like that.  :)

obsidianart:
on mac is cmd+click for chorme

blackforestboi:
[...]

PR #92, opened 2017-04-22
Expose search query in window location query.

Treora:
The UI logic should not have to bother about which pieces of state are exposed in the location query (in my opinion).
Therefore, this commit introduces a listener that bidirectionally syncs variables between the state and the window location.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "classnames": "^2.2.5",
     "diff-match-patch": "^1.0.0",
     "docuri": "^4.2.1",
+    "history": "^4.6.1",
     "lodash": "^4.17.4",
     "moment": "^2.18.1",
     "page-metadata-parser": "git+https://github.com/Treora/page-metadata-parser.git#npm-testable",
--- a/src/overview/enhancer.js
+++ b/src/overview/enhancer.js
@@ -1,0 +1,24 @@
+import { compose } from 'redux'
+
+import ReduxQuerySync from 'src/util/redux-query-sync'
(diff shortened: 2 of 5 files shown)


blackforestboi:
Hey @Treora 

I tried it out with `query-in-url`-branch and it's not working for me on chrome. 
Though I may expect something different than you actually did. 

#### What I thought would happen: 
When typing a query in the search field, the url in the address bar automatically changes to something like that:
`chrome-extension://ikhdjbhnddpfbfaompkhmlbhgchogffh/overview/overview.html?searchQuery=test`

It stays empty though.

Treora:
Did you build and reload the extension? It does work like that for me.
query-in-url

Treora:
Oh of course; the URL is not visible when viewing the overview as a new tab (I opened it with Ctrl+Y or the browser action button).

blackforestboi:
> Oh of course; the URL is not visible when viewing the overview as a new tab (I opened it with Ctrl+Y or the browser action button).

Yes, I just figured and wanted to add that to my description :) 
Is that a bug or restriction?

Treora:
> > Oh of course; the URL is not visible when viewing the overview as a new tab (I opened it with Ctrl+Y or the browser action button).
> 
> Yes, I just figured and wanted to add that to my description :)
> Is that a bug or restriction?

That's just what Chrome|ium does. It still works fine though. E.g. if you follow a link and go back, your query is still there (thus fixing your complaint in #52).

PR #95, opened 2017-04-29
Full Browser Bookmark Import ready

lewizho99:
The Bookmark Import is ready but not yet called similar to the History Import.

### DIFF
--- a/src/browser-bookmarks/background/import-bookmarks.js
+++ b/src/browser-bookmarks/background/import-bookmarks.js
@@ -1,0 +1,274 @@
+// Imports the full browser's bookmarks into our database.
+// The browser's bookmarkItems and visitItems are quite straightforwardly
+// converted to pageDocs and visitDocs (sorry for the confusingly similar name).
+
+
+import db from 'src/pouchdb'
+import { updatePageSearchIndex } from 'src/search/find-pages'
+import { isWorthRemembering,
+		 generateVisitDocId,
+         visitKeyPrefix,
+         convertVisitDocId } from 'src/activity-logger'
+import { generatePageDocId } from 'src/page-storage'
+
+
+/*
+@Description
+Array getBookmarkTree()
+Get the Whole Browser Bookmark Tree.
+
+@Params
+void
+
+@Return
+Array bookmarkItems  
+*/
+async function getBookmarkTree() {
+
+    const bookmarkTree = await browser.bookmarks.getTree();
+    const bookmarkItems = bookmarkTree[0].children[0].children.concat(bookmarkTree[0].children[1].children);
+    
+    return bookmarkItems.filter(({url}) => isWorthRemembering({url}));
+}
+
+
+/*
+@Description
+Object getRecentBookmarks(count = 5)
+Get Recent Browser Bookmark Items.
+
+@Params
+Integer count = 5
+
+@Return
+Object Obj.quantity 
+*/
+export async function getRecentBookmarks(count = 5) {
+    const recentBookmarkItems = await browser.bookmarks.getRecent(count)
+    return {
+        quantity: bookmarkItems.length,
+    }
+}
+
+
+/*async function getBookmarkTreeItems([...bookmarkTreeId] = []) {
+    const bookmarkTreeItems = await browser.bookmarks.get(bookmarkTreeId);
+    let   
+
+    for(let i = 0; i < bookmarkTreeItems.length; i++){
+
+
+    }
+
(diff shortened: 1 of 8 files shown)


Treora:
Hi, did you happen to coordinate this with @poltak? We've been discussing approaches in #29.

lewizho99:
@treora, i talked with @oliversauter a while ago and i got in touch with @poltak when i was almost through but for now the bookmark import is very analogous to the history import. That is it gets the bookmarks convert each bookmark item to a page doc and a visit doc then save into the db

Treora:
Sounds similar to #47.
@poltak and I have been discussing the overall approach to importing bookmarks and history. He is getting into this while I am trying to focus on other things, so he'll be the one to talk to.

lewizho99:
i am a bit confused here because i talked to @poltak before working on it and imidiately i finished and 
1), @poltak told me to talk to you @Treora  about this PR as he is busy with another module
2), so what happens to this PR?

poltak:
@lewizho99: I didn't say talk to him __because__ I am busy. I simply stated that he may also be interested in it. Right after I said that I'll give you some feedback on it (this PR) by tomorrow. 

I haven't forgotten about you, just gotta sit tight until I can get you some feedback on this 😉

lewizho99:
ah cool, hoping to hear from you soon... thanks for the quick response
hoping to hear from you.

src/browser-bookmarks/background/import-bookmarks.js
```diff
@@ -0,0 +1,256 @@
...
+Object getRecentBookmarks(count = 5)
+Get Recent Browser Bookmark Items.
+
+@Params
+Integer count = 5
+
+@Return
+Object Obj.quantity
+*/
+export async function getRecentBookmarks(count = 5) {
```
blackforestboi, line 42: Why only 5?

poltak:
@lewizho99: Thanks for helping out. First off, I'd recommend getting familiar with #29 and all the discussion in that thread. Most of the things I can see wrong in this PR so far are related to existing discussion in there. I'll put links to previous discussion in code comments.

The main issue with this so far is that we haven't decided on an actual structure for representing our bookmarks yet. Here I can see you're reusing the `page` and `visit` docs from `import-history`, but we were thinking of a third type of doc, `bookmark`, to use with `import-bookmarks` (discussion [here](https://github.com/WebMemex/webmemex-extension/issues/29#issuecomment-281030179), [here](https://github.com/WebMemex/webmemex-extension/issues/29#issuecomment-296225864)). As far as I can see, the logic in this PR is the same as `import-history`, just using a different source (`browser.bookmarks` vs `browser.history`).

Other minor things:

- The documentation comments seem non-standard (maybe from an IDE?); I think better just stick to standard JS comments above a function to explain anything unclear from code or just use [JSDoc](https://en.wikipedia.org/wiki/JSDoc) as it's so widespread and supported
- Try to keep out development stuff from commits (commented out stuff, `console.log` calls, code with comments above saying "ignore this"/"test", etc); git commit patch (`git commit -p`) is great for this, giving you an interactive mode to choose which lines to stage in a commit (or try a good graphical git client that allows you to easily stage only parts of your work, like [Tower](https://www.git-tower.com/))
- Try to give some more information with each commit; one easy way to do this is just stage a single atomic change per commit instead of everything (again `git commit -p` helps here), although not everyone works that way. If there's more commit information, I can more easily see your thought process and how your PR changes over time ([here's some good guidelines](https://chris.beams.io/posts/git-commit/#seven-rules))

src/browser-bookmarks/background/import-bookmarks.js
```diff
@@ -0,0 +1,256 @@
...
+Array getBookmarkTree()
+Get the Whole Browser Bookmark Tree.
+
+@Params
+void
+
+@Return
+Array bookmarkItems
+*/
+async function getBookmarkTree() {
```
poltak, line 23: I think it's better to just stick to using `browser.bookmarks.getRecent()`, as talked about [here](https://github.com/WebMemex/webmemex-extension/issues/29#issuecomment-283826516).

If `getRecentBookmarks` function is fixed up (it currently returns the count of found bookmarks, upper-bound of which is specified by the input...), I think you should replace all calls to `getBookmarkTree` with that.

Treora:
Closing due to inactivity. A more likely implementation for this feature is being made by @poltak over at WorldBrain (see https://github.com/WorldBrain/WebMemex/pull/25).

PR #97, opened 2017-05-01
Background/non-tab-dependent page fetch and analysis logic

poltak:
Mirrors [PR #15 on WorldBrain fork](https://github.com/WorldBrain/WebMemex/pull/15). In short it provides a fetch&analyse logic, as discussed in #28, allowing non-browser-tab dependent fetching and analyse logic. 

Creates new background `fetch-page-data` logic, which mirrors what happens when a user manually visits a page, but does it on a DOM gotten via XHR rather than DOM in-tab.

Updates calling functions in `page-analysis/background` and `page-storage/store-page` to default to the new background fetch&analysis if no `tabId` of a browser tab is supplied (tried to minimise interface changes; just make `tabId` optional).


src/page-analysis/background/fetch-page-data.js
```diff
@@ -0,0 +1,49 @@
+import extractPageContent from 'src/page-analysis/content_script/extract-page-content'
```
Treora, line 1: Importing a file from `.../contentscript/...` into a background script; apparently the file should move one directory up. (the separation into `content_script` and `background` dirs may not be ideal anyway, I'm open for other suggestions)

By the way, within the `page-analysis` 'submodule', a relative import can be used.

src/page-analysis/background/fetch-page-data.js
```diff
@@ -0,0 +1,49 @@
...
+const fetchDOMFromUrl = async url => new Promise((resolve, reject) => {
+    const req = new XMLHttpRequest()
+
+    req.onload = () => resolve(req.responseXML)
+    req.onerror = () => reject(req.statusText)
+
+    req.open('GET', url)
+
+    // Sets the responseXML to be of Document/DOM type
+    req.responseType = 'document'
```
Treora, line 49: I had not thought of using this option at all, brilliant.

src/page-analysis/background/get-fav-icon.js
```diff
@@ -1,20 +1,26 @@
+import favicon from 'favicon'
```
Treora: Does the favicon module fetch the whole page again to read the url of its icon? Should be an unnecessary round trip (at least with pages that don't cache by default) if we are fetching the page anyway.

Fetching the DOM once and reusing it would however require a bit of reorganisation in the analysis code. May be worth considering though; we had to do a similar thing for reusing a loaded pdf to get both the metadata and extracted text from it.

src/page-analysis/background/get-fav-icon.js
```diff
@@ -1,20 +1,26 @@
...
+
 import responseToDataURI from 'src/util/response-to-data-uri'
 
-// Get a tab's fav-icon (website logo) as a data URI
-async function getFavIcon({tabId}) {
-    const tab = await browser.tabs.get(tabId)
+// Async wrapper around favicon package
+const fetchFaviconUrl = url =>
+    new Promise((resolve, reject) =>
+        favicon(url, (err, favIconUrl) => err ? reject(err) : resolve(favIconUrl)))
```
Treora: You might like to use [pify](https://www.npmjs.com/package/pify) (promisify) to just say `pify(favicon)` instead.

src/page-analysis/background/index.js
```diff
@@ -46,11 +44,44 @@ async function performPageAnalysis({pageId, tabId}) {
...
+export async function analysePageInBackground({ page, url }) {
+    // Run page data fetching in background
+    const extractPageContent = () => fetchPageDataInBackground({ url })
+    const extractFavIcon = () => getFavIconFromUrl({ url })
+
+    await performPageAnalysis({ pageId: page._id, extractPageContent, extractFavIcon })
+    // Get and return the page.
+    const revisedPage = revisePageFields(await db.get(page._id))
+    return { page: revisedPage }
+}
```
Treora: It may be a messy if we keep trying to use the same `performPageAnalysis` function for different scenarios by passing in each analysis function as an argument. Looks like some refactoring could make things cleaner; not sure which factoring to use, would need to move things around a bit to see what works. I would start with inlining `performPageAnalysis` into the two `analysePageIn...` functions, then again pulling out the pieces they share.

Treora (commented):
I like the direction this is going. Also great you are adding better docs to functions than I have usually done.

As I also indicated in the code comments, since we are changing the current analysis flow to handle two scenarios (page in tab, page in background), we may want do a more thorough refactoring. I fear we end up with fragile code if we add many `if tabId` kind of statements in several places.

Not sure yet what a coherent design would look like. Could/Should we abstract the 'observed page' so most of the program flow can ignore whether it is in a tab or not?
Or should we share less code between the two analysis pipelines, and just make two separate functions for them? (or may these two approaches actually and up being the same?)

Maybe you already have an abstraction in mind that I did not perceive yet, I would be happy to hear.

src/page-analysis/background/get-fav-icon.js
poltak: Yes it does do another fetch. I was thinking how to get favicon if you're given a URL. Better to think how to get it if you're given the DOM, as we already have that in both contexts. Something like: `document.querySelector("link[rel*='icon']").href` should work if the page has a favicon.

poltak:
> since we are changing the current analysis flow to handle two scenarios [...], we may want do a more thorough refactoring. I fear we end up with fragile code if we add many if tabId kind of statements in several places.

Yeah I understand the concern, but I haven't thought of the best way yet. 

Looking over the entire two scenarios now, the only differences are:
- tab-based scenario stores screenshot; background does not ([`page-analysis/background/index.js#L46`](https://github.com/WorldBrain/WebMemex/blob/feature/fetch-and-analyse/src/page-analysis/background/index.js#L46))
- each scenario go down different fetch&analyse routes ([the very next line](https://github.com/WorldBrain/WebMemex/blob/feature/fetch-and-analyse/src/page-analysis/background/index.js#L47)). Both of these fetch&analyse call the same `extract-page-content` eventually, they just differ in the way they get the DOM/Document (from browser tab or over network via XHR)

Given they are the only differences I can think of, at the moment, I don't think we should try separating the code, but maybe just making different entry points? 

I cleaned up the conditional `tabId` code (and stopped passing in of different `extract-page-content` callbacks to the common fn) in `page-analysis/background` in 91813eb6f947e2a2cef5f57118092f41c22c6117. Now calling either entrypoint fn will result in the same thing, just using different methods. 

If you think that's on-the-right-track the next step would be to do the same with `page-storage/store-page` (one step "back" in the scenario flows, and currently acts at the entry point to all this logic -- from `log-page-visit` at least).

So for example, turn:
```
reidentifyOrStorePage: ({ url: string, tabId?: string }) =>
    Promise<{ page: IPage, fetchFinalPage: () => Promise<IPage> }>
```
into:
```
reidentifyOrStorePageInTab: ({ url: string, tabId: string }) =>
    Promise<{ page: IPage, fetchFinalPage: () => Promise<IPage> }>

reidentifyOrStorePageInBackground: ({ url: string }) => 
    Promise<{ page: IPage, fetchFinalPage: () => Promise<IPage> }>
```
Maybe should change their names as well...

src/page-storage/store-page.js
```diff
@@ -1,66 +1,97 @@
...
-            samePageCandidates,
-        })
-
-        // Return the page stub, and a promise of the analysed & deduped page.
-        return {page, finalPagePromise}
+        // Create a new page doc stub in the database.
+        const page = await storePage(createPage({ url }))
+
+        // Set up fetching function to run anaylsis + fetching logic on-demand
+        const fetchFinalPage = () => analyseAndTryDedupePage({ tabId, page, samePageCandidates })
```
Treora, line 60: So your idea is that analysis is not invoked here but by the caller? So the task of `reindentifyOrStorePage` would then only be to create *some* page document. Might be good, not sure.

If we do this change, we may then not even have to prepare and return this function; the caller could call analyse&dedupe directly (the caller knows `tabId`, and gets passed the `page` object, and `samePageCandidates` could be recomputed anyhow). The caller may just want to know whether the page needs analysis; we could return a boolean (e.g. `isReused`), or (probably better) it could just figure it out from the page contents.

src/page-storage/store-page.js
```diff
@@ -1,66 +1,97 @@
...
     }
+
+    // Do the data XHR fetch + analyse
+    const { text: extractedText, metadata: extractedMetadata, favIcon } = await fetchPageData({ url })
+
+    // Create and store DB page using data, before returning it
+    const page = createPage({ url, extractedText, extractedMetadata, favIcon })
+
+    // Store an augmented version of the doc, using data derived from extracted page data
+    const augPage = await storePage(revisePageFields(page))
```
Treora, line 91: The idea of `revisePageFields` is to run such changes at retrieval time, for flexibility (e.g. to transform fields of older docs without needing a database migration). So no need for it here, I suppose.

src/page-storage/store-page.js
```diff
@@ -1,66 +1,97 @@
...
+    // Find pages we know that had the same URL.
+    const samePageCandidates = (await findPagesByUrl({ url })).rows.map(row => row.doc)
+
+    // Check if we can already tell in advance that this is the same page.
+    const reusablePage = await tryReidentifyPage({ url, samePageCandidates })
+
+    // Equality is known (or assumed) in advance. Reuse the old page as is.
+    if (reusablePage) {
+        return reusablePage
     }
```
Treora, line 82: Can we reuse `reindetifyOrStorePage` instead of duplicating the above lines?
It seems this whole function follows the same pattern as in the case storing a page from a tab; I suspect we should be able to make a single pipeline with a different implementation of the actual (fetch &) analysis (which may internally again use the same extraction functions). I made a [sketch](http://penpub.ink/page47.html) of how things fit in my mind. It does not answer the question how the data is stored. I like to have each finished piece of analysis be stored to database directly, for which it is easiest to store them inside the analysis function instead of returning the completed page object when ready. This is the current approach. It would also be clean if the page-analysis module does not know about the database, but it would then be hard to still incrementally store the analysis parts. Trade-offs..

poltak: So a few changes have happened over time that led to creating this `storePageFromUrl` fn. Previously I was using the `reindetifyOrStorePage` as the batcher function. However this led to numerous changes that propagated over several files, like making functions' `tabId` arg optional or creating separate function entrypoints to handle the two scenarios (`url` vs `tabId`). That was okay, but the real issue was `reindetifyOrStorePage` does everything needed (reidentifying, analysis, deduping, storage) but it does it in two stages (stage 1 is reidentifying + storing page stub, 2 is analysis, dedup + final page storage via the `finalPagePromise`). This seems to work nicely for page visits, but for the batcher it isn't ideal, and required some hacky extensions to be able to do things like pull out the inner promise (stage 2) to batch, as that's the more intensive stuff that we want to batch.

Originally solved this by doing what I wrote about before and change the `finalPagePromise` to an async function `fetchFinalPage`, to essentially defer it until invocation (unlike a Promise which starts to resolve on instantiation). The problem with this way was the batcher needed to still resolve the initial returned Promise for `reindetifyOrStorePage` (stage 1) for each input at the same time, to get that inner async function (stage 2) which it can defer and batch on (then continue to process _n_ number of them at a time). That means the space complexity for the imports process is dependent on input size, but the whole point of the batcher is to constrict space complexity to a constant space (that _n_ level of concurrency * constant space needed to process an input), so we can act on large inputs like browser history without worrying about OOMing.

So that led to the creation of this `storePageFromUrl` which essentially does everything that `reindetifyOrStorePage` does, but in a single stage/promise. As it has the intention of running in the batcher we can also remove parts that aren't essential to run for each input, like reindexing (leave till batch completion) and the whole page stub thing (the batcher treats each invocation of the function on each input as an atomic transaction; it will either finish and store a new page or cancel and nothing is stored - cancel from error won't be re-attempted, but cancel from batcher interruption like pause can be reattempted at a later batcher restart).


__tl;dr__
Because of some differences in how all this page logic (reidentifying, analysis, deduping, storage) is better suited for use in different scenarios (run once on user page-visit vs run many times in a large batch), having `storePageFromUrl`, which optimises the processing flow for batch use (and re-uses existing logic, or reimplements it, if it's really necessary), seems better than modifying the existing logic too much as it already seems appropriate for the page-visit scenario.

If that makes sense and you agree, there's no need for me to keep the changes to `reindetifyOrStorePage` and that can go on existing as it did before, as it's only used in the page visit scenario.

src/page-storage/store-page.js
poltak: This is here as it is done in the `reindetifyOrStorePage` scenario (it eventually calls `analysePageInTab` in the stage 2 `finalPagePromise` analysis logic which calls `revisePageFields` before retuning an analysed, revised page). I was basically flattening out the callstack that `reindetifyOrStorePage` generates (and omitting unnecessary things), but may split this across different modules like `reindetifyOrStorePage` if needed.

src/page-storage/store-page.js
poltak: To give you more of an idea of the batcher in use with `storePageFromUrl `, I have played with rewriting the existing `import-history` to be based on the batcher in [this commit](https://github.com/WorldBrain/WebMemex/commit/520074884ac87334137fdcc2b2c7a7500d7aaf8e) on WorldBrain branch where I'm working on hooking things up.

It's not final but I'm planning to init the batcher here in the background and make some interface to it, via the RPC wrapper lib that we have, to call and manage from the UI redux actions. Main things is the observer functions (`next` run after each input processed, `error` run after any errors cancel out the input processing, and `complete` run after all input is processed or cancelled). You can see reindexing, and visit page generation and storage is run after the page batch complete (visits are derived from pages, and needs all pages to do the inter-visit doc referencing).

Treora: You may be right that `reidentifyOrStorePage` should not start the next stage. Cleaner would perhaps be to have a single controlling function (like sketched) that invokes each of the steps (up to 4 steps, if dedupe and reindexing are counted separately). Don't feel the need to keep things as the existed before; it requires refactoring to keep things maintainable.

I see how trying to delay/batch a part of the process could change the requirements. I may be wrong, but I would still expect a closer similarity though. Perhaps we would have two slightly different functions that walk through the steps; making them look similar would make it easier for us to update them both when new features come along. Maybe they could even be one function with an option (e.g. whether to reindex).

> You can see reindexing, and visit page generation and storage is run after the page batch complete (visits are derived from pages, and needs all pages to do the inter-visit doc referencing).

Delaying reindexing makes sense. Possibly deduplication could also wait, but may be better to do directly.
Doing the visits after having processed all the pages does not sound completely right to me. I would probably still opt for importing and creating (nearly empty) page docs first, and add their visits directly; then in the batching process we create new pages for the downloaded docs and possibly letting deduplication clean them up if desired, as discussed in #29. Or does this sound overcomplicated to you?
(edit) Or even in case we would want to fill up the nearly empty page docs rather than creating a new one, we could first run `reidentifyOrStorePage` for each url to create stubs, create all the visits, and then start the batch process to do the fetch&analysis to fill in each page stub.

poltak: The thing I don't get is you have to update the references between visit docs (changing the `referringVisitId` gotten from browser model to an `_id` of another visit doc in our model). This needs all the visit docs available to do so, hence needs to happen right at the end before reindexing. Could still store visit docs directly at each step, but then the reference updating still needs to happen on completion.

>  I would probably still opt for importing and creating (nearly empty) page docs first

We could stick with this way, and batch them in another batching instance, if needed, but I just don't understand the purpose behind the page stubs. In page visit scenario, for a user it means the pages will show up in overview before all the data fetch/analyse stuff is done after the user visits the page (is that correct?), but in the importing history/bookmarks scenarios, what is the purpose?

Treora: Sorry for the delay, I thought I replied already. The referrer references between the visits (trails!) are valuable information to store. This in practice indeed requires to have the visit docs made to know their `_id`, though if we would process history in temporal order items we should be able to link the visits up while importing. In any case importing visits does not require all pages to be fetched and analysed. I would consider not trying to batch the actual browser history import (which takes seconds at most), but only the fetch&analysis of pages after that.

Creating page stubs is part of importing the browser history. I see the act of fetching&analysing pages as a separate process. One might want to import the whole history, but fetch only a part of it. The URL and title of the page (and visit time and trail) are valuable things to remember, and possibly the only thing in case the page cannot be fetched anymore. It makes most sense to me to not rely too much on the fetching process; e.g. to allow deferring the fetching to another time.

poltak: No rush with replies. 

Alright so by breaking it into separate "browser history import" and "fetch&analysing pages" processes, in your mind, what does each process consist of? Is "browser history import" just the `browser.history.search()` call to get `HistoryItem`s, along with the creation of `page` stub docs for each of them? If so, then yeah the visit docs should be able to be done in that stage (cannot do in batch stage as processing order is not guaranteed, by design).

> It makes most sense to me to not rely too much on the fetching process; e.g. to allow deferring the fetching to another time.

This makes me question when these processes/stages will happen in relation to user actions in the UI. At the moment the UI is based on @oliversauter's mockups, linked to in #65 (implementation now looks like [this](http://imgur.com/a/EXS1b)), and pressing "Start Import" currently starts off the batcher (with the other buttons all essentially affording user-control of the batcher's state), which is already ready-to-run in the background with input URLs gotten from a `browser.history.search()` call on background script start. 

As you have noted, history import process shouldn't take long (compared with batch fetching process), but it's input size dependent, so probably should not just happen on background script load, as I'm playing around with now. It could happen on the "Start Import" press, or even when the user clicks on the "Import" option in options nav, or even have a pre-import view before the estimates view that runs that process when the user wants to proceed into import estimate view. As you brought up the possibility of allowing fetching process to be deferred, did you already have something in mind for when the "browser history import" process should happen relative to UI/user actions?

Treora:
(pulling this general conversation out of the [in-line discussion](https://github.com/WebMemex/webmemex-extension/pull/97#discussion_r116375854) as that may become hidden at some moment)

> in your mind, what does each process consist of?

Browser history import would do what the [import-history](https://github.com/WebMemex/webmemex-extension/blob/edf2252475ee3358021314500051ca3d4b7c2f00/src/browser-history/background/import-history.js) I drafted does: import browser history, convert and store it as pages and visits. The batch-fetch process would download docs and also add them as pages (now with more content) to the database. Either can be invoked without the other (we could pass the batch-fetch process any list of URLs).

Note this all just concerns the technical design. The UI could hide the separation, and run the history import and start bath fetching in direct succession.

I imagine that starting batch fetching could mean just adding a bunch of URLs to a list in the database, which the background process reads and processes as a queue whenever it feels like (possibly only during inactive & power-plugged moments).

Treora:
Closing this old PR. This might actually be an interesting feature still, but it would have to redesigned&rewritten if it gets attention again.

PR #102, opened 2017-05-29
Support deleting visits/pages (fix #21)

### LINKED ISSUE #21: Forget individual pages
One should be able to remove individual pages/visits from their memory. At least from the history/search overview, through a context menu, or (fancier) by dragging it to a bin. Perhaps also through a menu on the to-be-forgotten page itself.

When we have implemented a blacklist (#20), we can also ask whether to enlist the domain of the removed page and/or forget all pages from that domain.

See also https://github.com/WorldBrain/Research-Engine/issues/46.

Treora:
Clicking the trash bin button now deletes that specific visit. It also
deletes the page object the visits refers to; unless there is still another
visit referring to it, or to an 'equivalent page': i.e. a page redirecting
(possibly indirectly) to the same page through a 'seeInstead' link.

Fixes #21.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "pouchdb-browser": "^6.1.0",
     "pouchdb-find": "^0.10.4",
     "pouchdb-quick-search": "^1.2.0",
-    "pouchdb-upsert": "^2.0.2",
     "prop-types": "^15.5.10",
     "react": "^15.4.1",
     "react-datepicker": "^0.43.0",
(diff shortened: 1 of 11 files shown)


blackforestboi:
I wonder if the use case is rather having one visit or all traces of the page deleted? 
My gut says more the latter, but it may be an option to have a prompt, where the user is asked if he either wants to delete just the visit or everything about this page?

Treora:
I decided to take the most obvious approach: clicking the trash button on a visit deletes that visit. Not all visits to that same page object, nor to all pages from the same URL, nor to the same domain. This seems fine for MVP, other options can come later.

Issue #103, opened 2017-05-30
Implementation of a Testing Suite and Continuous Integration Pipeline


reficul31:
This is the testing project for Google Summer of Code 2017. This project aims to setup a testing suite and continuous integration pipeline for the webmemex extension.

For the project, the testing team has decided to use the [Jest](https://facebook.github.io/jest/) framework written by facebook. To have a look at the mock testing suite([branch](https://github.com/reficul31/webmemex-extension/tree/jest)) build during the community bonding period.

## Issue List
- [x] **Issue 1** -- Integration with the Linter (May 30-31)
- [ ] **Issue 2** -- Unit tests for the Activity Logger module (Jun 1)
- [x] **Issue 3** -- Unit tests for the Freeze Dry module (Jun 2-4)
- [ ] **Issue 4** -- Systems tests for Options module (Jun 5)
- [ ] **Issue 5** -- Snapshot tests and other basic tests for the Overview module (Jun 6-7)
- [ ] **Issue 6** -- Unit tests for the util functions (Jun 8-9)
- [ ] **Issue 7** -- Page analysis module unit tests (Jun 10-11)
- [ ] **Issue 8** -- Page viewer module unit tests (Jun 12-13)
- [ ] **Issue 9** -- Integration with TravisCI (Jun 14-17)

**Student:**  
Shivang Bharadwaj / [@reficul31](https://github.com/reficul31)

**Mentors:**  
Jonas Aschenbrenner / [@sanjo](https://github.com/sanjo )  
Dan Lüdtke / [@danrl](https://github.com/danrl)  
Stefano Solinas / [@obsidianart](https://github.com/obsidianart)  
Raj Maurya / [@raj-maurya](https://github.com/raj-maurya)

**Proposal:**[Link](https://docs.google.com/document/d/11qTCBBGhyY6-_0ZuMXs2kRCz0EYimLOc5EM51uY6h8I/edit)  
**Blog:**[Updates](http://shivangbharadwaj.com/blog)  
**Issue List:**[Issues](https://docs.google.com/document/d/1JtZt1V5QG1dwCdUhXpaDEPlDyy0pZURC-t9uyvoe8v8/edit?usp=sharing)  
**Branch:**[Work](https://github.com/reficul31/webmemex-extension/tree/project)
**WiP:**[Pull Request](https://github.com/WebMemex/webmemex-extension/pull/107)

Treora:
[...]

PR #104, opened 2017-05-30
Integration of Linter with Jest Presets

reficul31:
[Refs](https://github.com/WebMemex/webmemex-extension/issues/103) Issue 1 and Issue 2. The PR consists of the integration of the Linter with the jest presets and also the unit tests for the activity logger module.
This PR also sets up the basic testing suite with jest.  
Not sure whether the linter should watch the test files as well as the src files, therefore I have left that part blank.


Treora:
Nice. Jest looks good to me, had not heard of it before. To simplify discussion, it would be easier to put the tests themselves in a separate PR. I'll just comment on the jest setup itself now.

.eslintrc.json
```diff
@@ -1,16 +1,21 @@
...
         "plugin:import/recommended",
         "plugin:import/react",
         "plugin:promise/recommended",
         "plugin:react/recommended",
         "standard",
         "standard-react"
     ],
+    "globals": {
+        "db": true
+    },
```
Treora: `db` should not be a global. It is exposed as a global for console debugging, but using it as a global anywhere in the code is a dangerous mistake (it will break in production).

.eslintrc.json
```diff
@@ -54,6 +60,8 @@
         "react/jsx-filename-extension": "error",
         "react/prefer-stateless-function": "warn",
         "react/sort-comp": "warn",
-        "react/jsx-max-props-per-line": [1, { "when": "multiline" }]
+        "react/jsx-max-props-per-line": [1, { "when": "multiline" }],
+        "import/namespace": "off",
+        "import/default": "off"
```
Treora: Are these lines related to the jest setup, or have another reason?

.eslintrc.json
```diff
@@ -1,16 +1,21 @@
 {
```
Treora: If we keep tests in their separate folder, we could add a separate eslint config there, which extends this one and adds the jest plugin&settings (see eg. this [QA](https://stackoverflow.com/questions/38615872/configure-different-eslint-rules-for-a-set-of-paths#38770465)).

package.json
```diff
@@ -48,27 +48,37 @@
...
     "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-es2015": "^6.18.0",
-    "babel-preset-react": "^6.16.0",
+    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-react": "^6.24.1",
     "babel-preset-stage-1": "^6.16.0",
     "babelify": "^7.3.0",
     "browserify": "^13.1.1",
     "css-modulesify": "^0.27.1",
+    "enzyme": "^2.8.2",
```
Treora, line 59: Is this used?

package.json
```diff
@@ -48,27 +48,37 @@
...
     "gulp-uglify": "^2.0.1",
+    "identity-obj-proxy": "^3.0.0",
+    "jest": "^20.0.1",
+    "jest-cli": "^20.0.0",
     "loose-envify": "^1.3.0",
     "postcss-cssnext": "^2.9.0",
+    "pouchdb": "^6.2.0",
+    "pouchdb-adapter-memory": "^6.2.0",
+    "react-addons-test-utils": "^15.5.1",
+    "react-test-renderer": "^15.5.4",
```
Treora: Are the react tools used? Could add them already of course if you know you'll need them later; or they could be added when needed to make it easier to understand the commit logs (e.g. git blame tells why a module was introduced). I may be nit-picking here though.

Treora: Are the react tools used already? Could add them already of course if you know you'll need them later; or they could be added when needed to make it easier to understand the commit logs (e.g. git blame tells why a module was introduced). I may be nit-picking here though.

Treora: Are the react tools used already? Could add them already of course if you know you'll need them later; or they could be added when needed to make it easier to understand the commit logs (e.g. git blame tells why a module was introduced). I may be nit-picking here though.

reficul31:
> To simplify discussion, it would be easier to put the tests themselves in a separate PR.

I thought a couple of testing suites will help visualize how the whole setup will turn out. I also needed a few tests to check if the environment was functioning the way it was supposed to. Although if it is a causing ambiguity as to the purpose of the PR I will be sure to break it up into two PRs.

Treora (commented):
I think we can do without the mock for pouchdb, it would be bothersome to have to keep the two files in sync.
Not sure what's the cleanest way. Adding a test for `env='test'` in `src/pouchdb.js` is one option.

Another thing, do you have a preference for having all tests in one folder, as you have done now, or would you think putting `.test.js` files besides each file is a nicer approach? I have often seen the latter, it makes a larger mess in folders but avoids having to create and maintain the same folder structure twice, and makes it easier for developers to quickly open the test for a file.

.eslintrc.json
reficul31: Yes these lines are some of the recurring errors that occurred during the creation of the [mock testing suite](https://github.com/reficul31/webmemex-extension/tree/jest)

package.json
reficul31: Yes this was a part of the React Testing tool used while snapshot testing different components. You can see its implementation in the [mock testing suite](https://github.com/reficul31/webmemex-extension/tree/jest).

package.json
reficul31: Yes they were a part of the mock testing suite. I can remove them if it causes confusion. It's all the same to me if we add it here or after,but I know they will be used for sure. Again, I know I am being quite repetitive, refer [mock testing suite](https://github.com/reficul31/webmemex-extension/tree/jest)

reficul31:
> I think we can do without the mock for pouchdb, it would be bothersome to have to keep the two files in sync.
Not sure what's the cleanest way. Adding a test for env='test' in src/pouchdb.js is one option.

@sanjo

> Another thing, do you have a preference for having all tests in one folder, as you have done now, or would you think putting .test.js files besides each file is a nicer approach? I have often seen the latter, it makes a larger mess in folders but avoids having to create and maintain the same folder structure twice, and makes it easier for developers to quickly open the test for a file.

If it were totally up to me, I would use the different folder approach as I feel it keep the code quite clean and segregated. But hey, that's just me. I feel that we should ask my mentors for their advice on this topic.
@sanjo @obsidianart @danrl @raj-maurya

.eslintrc.json
reficul31: I thought so too at first. But if you refer to [line](https://github.com/WebMemex/webmemex-extension/pull/104/files#diff-cd67c91cd27eef6025852e6802ed2512R5) , it takes advantage of the jest mocking feature and uses db as a global variable. Removing it from the ```globals``` causes it to throw an error stating db is note defined.

Treora: Ok, not sure how the mock system works; I would find it weird if it adds such globals implicitly. Anyway, you could add a line `/* global db */` to the file to suppress the error (see [docs](http://eslint.org/docs/user-guide/configuring#specifying-globals)). Or it should be in a config only used for tests, because it is important to have this warning enabled in the rest of the code.

package.json
Treora: Do either way, as you like, I'm just nit-picking unimportant things.

reficul31:
Changes made 
- Removed globals from .eslintrc
- Added the comment to suppress the error
- Rearranged the commits to make more sense

ghost:
> I think we can do without the mock for pouchdb, it would be bothersome to have to keep the two files in sync.
Not sure what's the cleanest way. Adding a test for env='test' in src/pouchdb.js is one option.

Let's do this. Jest will set NODE_ENV automatically to 'test' ([Reference](https://facebook.github.io/jest/docs/getting-started.html)).

> Another thing, do you have a preference for having all tests in one folder, as you have done now, or would you think putting .test.js files besides each file is a nicer approach? I have often seen the latter, it makes a larger mess in folders but avoids having to create and maintain the same folder structure twice, and makes it easier for developers to quickly open the test for a file.

I prefer to have unit test files next to the implementation files because it makes it significantly easier to open them when you work on a module.

reficul31:
I have moved the test files to the src folders, however I can't seem to figure out how to check the environment of node and change the configuration of the adapter. Currently I am trying to do the following.

```
let db = new PouchDB({
    name: 'webmemex',
    auto_compaction: true,
})

if (process.env.NODE_ENV === 'test') {
    db = new PouchDB({
        name: 'testdb',
        auto_compaction: true,
        adapter: 'memory',
    })
}
```
I know the logic is extremely flawed and I would really appreciate some help.

ghost:
The check should be correct. Do you get any error? You can also probably use console.log to check, if it works correctly.

```js
let db
if (process.env.NODE_ENV === 'test') {
    db = new PouchDB({
        name: 'testdb',
        auto_compaction: true,
        adapter: 'memory',
    })
} else {
    db = new PouchDB({
        name: 'webmemex',
        auto_compaction: true,
    })
}
```

reficul31:
Using ```let db``` throws a linting error stating ```Exporting mutable 'let' binding, use 'const' instead``` . Currently I have suppressed the error, but I would really like a better way to do this.

Treora:
> Using let db throws a linting error stating Exporting mutable 'let' binding, use 'const' instead .

Can always work around that if needed:
```
let db_
...
const db = db_
```

Or maybe `export {db_ as db}` would also work?

Or one can do something like:
```
const db = (process.env.NODE_ENV === 'test')
    ? ....
    : ....
```

By the way, it would be nice if the extension can keep importing only `pouchdb-browser`, not whole `pouchdb` (=`pouchdb-browser`+`pouchdb-node`), to reduce bundle size. The test version could probably just use `pouchdb-node`, actually, and `require` it to import it conditionally.

Maybe something like this?:

```
import PouchDB from 'pouchdb-browser'

const pouchdbOptions = {
    name: 'webmemex',
    autocomplete: true,
}

function initDB() {
    return PouchDB(pouchdbOptions)
}

function initTestDB() {
    const PouchDB = require('pouchdb-node').default
    return PouchDB({
        ...options,
        name: 'testdb',
        adapter: 'memory',
    })
}

export const db = (process.env.NODE_ENV === 'test')
    ? initTestDB()
    : initDB()
```

reficul31:
Changes made
- Implementation of the db configuration based upon @Treora suggestion
- Removed some dependencies that weren't required

ghost:
We might also update the contribution guidelines to require or at least recommend that developers write unit tests for the code they add or modify.

reficul31:
@Treora I think the PR is ready to merge :D

src/pouchdb.js
```diff
@@ -7,13 +7,28 @@ import PouchDBFind from 'pouchdb-find'
 PouchDB.plugin(PouchDBQuickSearch)
 PouchDB.plugin(PouchDBFind)
 
-const db = new PouchDB({
+const pouchdbOptions = {
     name: 'webmemex',
-    auto_compaction: true,
-})
+    autocomplete: true,
```
Treora: That should be `auto_compaction`. My mistake when I sketched the code example.

src/pouchdb.js
```diff
@@ -7,13 +7,28 @@ import PouchDBFind from 'pouchdb-find'
...
-export default db
+function initDB() {
+    return PouchDB(pouchdbOptions)
+}
 
+function initTestDB() {
+    const PouchDB = require('pouchdb-node')
+    return PouchDB({
+        name: 'testdb',
+        auto_compaction: true,
```
Treora: I'd prefer 
```
...pouchdbOptions,
name: 'testdb',
```
so that any changes in the options would not have to be repeated here.

.eslintrc.json
Treora: If these are necessary for jest (ideally they would not), I would actually like to have still these useful warnings enabled for the rest of the code. Could we have a separate linter configuration file that is used for the test files? Might be nice anyway, as e.g. jest globals are not acceptable in source code. The test config could extend the normal config and add all the jest-specific things. Then we specificy `**/*.test.js*` in the arguments to eslint while pointing it at the new config file.

package.json
```diff
@@ -47,27 +47,37 @@
   },
   "devDependencies": {
     "autoprefixer": "^6.7.5",
+    "babel-jest": "^20.0.1",
```
Treora, line 50: Unneeded?

> Note: babel-jest is automatically installed when installing Jest

[(here)](https://facebook.github.io/jest/docs/getting-started.html#selector(exact=Note:%20babel-jest%20is%20automatically%20installed%20when%20installing%20Jest,type=TextQuoteSelector))

package.json
```diff
@@ -78,5 +88,33 @@
...
+    "moduleFileExtensions": [
+      "js",
+      "jsx"
+    ],
+    "unmockedModulePathPatterns": [
+      "<rootDir>/node_modules/react",
+      "<rootDir>/node_modules/react-dom",
+      "<rootDir>/node_modules/react-addons-test-utils",
+      "<rootDir>/node_modules/fbjs"
+    ]
```
Treora: I'd keep this config in a separate file, say `.jest-config.json`, and pass that with `--config <path/to/json>` ([docs](https://facebook.github.io/jest/docs/configuration.html#selector(exact=--config%20%3Cpath/to/json%3E,type=TextQuoteSelector)))

.eslintrc.json
reficul31: If we are making use of shareable configs, might I suggest my original approach of moving all the tests to their own folder and adding a folder config of the linter. It would make the approach much more cleaner and, in my opinion, make the code base more neat(for lack of a better word). I experimented with using extending the linter config, leading me to write this messy piece of code.

```
gulp.task('lint-test', () => {
    return gulp.src(['src/**/*.test.js', 'src/**/*.test.js'])
        .pipe(eslint({
            configFile: '.test-eslintrc.json'
        }))
        .pipe(eslint.format())
})

gulp.task('lint', ['lint-test'], () => {
...
}
```

reficul31: Also not to mention the presence of two eslint config files in the root of the code base. But again, this is just my opinion.

ghost: I would go for one eslint config. Otherwise you need to run eslint twice I think. E.g. CLI tool, editor integrations. I can only set one eslint config in Webstorm for example.

With the jest/globals there are two options:
1. Keep it in the eslint config and have the tradeoff that the testing globals are not reported as undefined in non-testing files.
2. Remove it in the eslint config and add `/* eslint-env jest */` in all test files.

I use option 2 in the main project that I'm working on.

For the import rules I don't really see why those need to be different for test files. Maybe we can resolve the underlying issue.

Treora: I am equally fine with Sanjo's approach, if we can fix the warnings.

In case we do need separate configs, I think I would prefer a few lines of messy gulp code over having to keep two folder trees in sync.

reficul31: Another problem that seems to occur with two config files and using gulp-eslint to run the linter is as follows. So basically I was using two config files and 2 tasks to run the linter for the test file and the normal src files. The problem arises when I run the linter for the src files, it also lints the test files as the test files are of the form *.js. I know a workaround but it will require the the use of [gulp-ignore](https://www.npmjs.com/package/gulp-ignore). Or we could use the 2nd approach suggested by @Sanjo .

package.json
reficul31: I reinstalled jest. I don't think it was installed automatically.

reficul31:
So after a lot of thought, I thought of going with second suggestion given by @Sanjo . Don't really have a reason, but mainly I was actively trying to avoid another dependency, in this case if I went with two configs that dependency would have been gulp-ignore.

reficul31:
@Treora any more thoughts on the PR?

src/pouchdb.js
Treora: If you could still do this change, I think it's ready to merge!

src/pouchdb.js
```diff
@@ -7,13 +7,29 @@ import PouchDBFind from 'pouchdb-find'
...
+function initDB() {
+    return PouchDB(pouchdbOptions)
+}
 
+function initTestDB() {
+    const PouchDB = require('pouchdb-node')
+    return PouchDB({
+        ...pouchdbOptions,
+        name: 'testdb',
+        auto_compaction: true,
```
Treora: `auto_compaction: true` should be `adapter: 'memory'`, no?

Treora: apart from this, looks ready to merge!

reficul31:
Changes made:
- Removed all the un-necessary rules for the testing suite.
- Added a jest config file that is being passed as --config
- Changed the config of testdb

blackforestboi:
Yeah! congrats @reficul31 :)

I cannot tell if its a good job, but I assume, since you passed Gerben's requirements :P

PR #106, opened 2017-06-07
Infinite scroll

Treora:
Building on from #105 (search by filtering), this loads more results when reaching the bottom of the visible list. It seems to work, apart from small bugs, however the code has become messy. It still need a fresh look and a bit of refactoring.

@poltak: I took a little inspiration from [your work](https://github.com/WorldBrain/WebMemex/pull/29) for WorldBrain, but chose to take a different approach. To make it work with the filter-based search, instead of keeping count of the number of rows displayed, I keep track of the last row that we know about.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "react-dom": "^15.4.1",
     "react-redux": "^5.0.1",
     "react-router": "^3.0.2",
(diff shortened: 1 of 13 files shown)


Treora:
I rebased and merged this (outside github) in a74943a.

PR #108, opened 2017-06-16
Unit tests for Freeze Dry Module

reficul31:
The tests are not final. A few inputs on the testing methods would be useful. Also jest-mock-fetch is a really good library, given the choice I would like to keep it as a dependency. Saying that I know it messes up the repository with one ugly file containing just a single line, but hey I am open to other approaches.

One of the tests is failing. Specifically the fix-links tests to check the base element was added to the head or not. Any input on why it is failing would be really helpful. Otherwise a general review of the code is needed badly.

### DIFF
--- a/src/freeze-dry/common.test.js
+++ b/src/freeze-dry/common.test.js
@@ -1,0 +1,104 @@
+/* eslint-env jest */
+
+import { inlineUrlsInAttributes, urlToDataUri, removeNode } from 'src/freeze-dry/common'
+import * as responseToDataUri from 'src/util/response-to-data-uri'
+import { dataURLToBlob } from 'blob-util'
+
+const imageDataUri = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAAAAAA6fptVAAAACklEQVR4nGNiAAAABgADNjd8qAAAAABJRU5ErkJggg=='
+
+beforeEach(() => {
+    fetch.resetMocks()
+})
+
+describe('removeNode', () => {
+    test('should remove the node', () => {
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><head></head><body></body></html>',
+            'text/html'
+        )
+        removeNode(rootElement.querySelector('head'))
+        expect(rootElement.querySelector('head')).toBeNull()
+    })
+})
+
+describe('urlToDataUri', () => {
+    test('should return a valid dataUri for a URL', async () => {
+        const spy = jest.spyOn(responseToDataUri, 'default')
+        await urlToDataUri('https://example.com/page')
+        expect(spy).toHaveBeenCalled()
+        spy.mockRestore()
+    })
+
+    test('should return a "about:invalid" upon failure', async () => {
+        expect.assertions(2)
+        const spy = jest.spyOn(responseToDataUri, 'default').mockImplementation(() => {
+            throw new Error('mock error')
+        })
+        const dataUri = await urlToDataUri('http://example.com')
+        expect(spy).toHaveBeenCalled()
+        expect(dataUri).toBe('about:invalid')
+        spy.mockRestore()
+    })
+
+    test('should return a "about:invalid" when fetching fails', async () => {
+        fetch.mockReject()
+        const dataUri = await urlToDataUri('http://example.com')
+        expect(dataUri).toBe('about:invalid')
+    })
+})
+
+describe('inlineUrlsInAttributes', () => {
+    const docUrl = 'https://example.com/page'
+    const parser = new DOMParser()
+    let imageBlob
+
+    beforeAll(async () => {
+        imageBlob = await dataURLToBlob(imageDataUri)
+    })
+
+    test('should change the URL in <img> tag to a dataUri', async () => {
+        fetch.mockResponseOnce(imageBlob)
+        expect.assertions(2)
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="public/image/background.png" alt="background" /></body></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'img', attributes: 'src', rootElement, docUrl})
+        expect(rootElement.querySelector('img').getAttribute('data-original-src')).toBe('public/image/background.png')
+        expect(rootElement.querySelector('img').getAttribute('src')).toBe(imageDataUri)
+    })
+
+    test('should change the URL in the <link> tag to a dataUri', async () => {
+        fetch.mockResponseOnce(imageBlob)
+        expect.assertions(2)
+        const rootElement = parser.parseFromString(
+            '<html><head><link rel="icon" href="public/image/favicon.ico"></head></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'link', attributes: 'href', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('link').getAttribute('data-original-href')).toBe('public/image/favicon.ico')
+        expect(rootElement.querySelector('link').getAttribute('href')).toBe(imageDataUri)
+    })
+
+    test('should change the URL in srcset of the <img> tag to a dataUri', async () => {
+        fetch.mockResponseOnce(imageBlob)
+        expect.assertions(2)
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="background.png" alt="background" srcset="background.png 200w,background.png 400w"></body></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'img', attributes: 'srcset', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('img').getAttribute('data-original-srcset')).toBe('background.png 200w,background.png 400w')
+        expect(rootElement.querySelector('img').getAttribute('srcset')).toBe(imageDataUri)
+    })
+
+    test('should remove the attribute integrity from the tag', async () => {
+        const rootElement = parser.parseFromString(
+            '<html><head><link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css" rel="stylesheet" integrity="sha256-MfvZlkHCEqatNoGiOXveE8FIwMzZg4W85qfrfIFBfYc= sha512-dTfge/zgoMYpP7QbHy4gWMEGsbsdZeCXz7irItjcC3sPUFtf0kuFbDz/ixG7ArTxmDjLXDmezHubeNikyKGVyQ==" crossorigin="anonymous"></head></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'link', attributes: 'href', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('link').getAttribute('integrity')).toBeNull()
+    })
+})
--- a/src/freeze-dry/fix-links.test.js
+++ b/src/freeze-dry/fix-links.test.js
@@ -1,0 +1,53 @@
+/* eslint-env jest */
+
+import fixLinks from 'src/freeze-dry/fix-links'
+
+describe('fixLinks', () => {
+    const docUrl = 'https://example.com/page'
+
+    test('should call the function to prepend the base element', async () => {
+        const rootElement = window.document.implementation.createHTMLDocument()
+        rootElement.querySelector('head').insertAdjacentElement = jest.fn()
+        const docUrl = 'about:blank'
+        await fixLinks({rootElement, docUrl})
+        const base = document.createElement('base')
+        base.href = docUrl
+        expect(rootElement.querySelector('head').insertAdjacentElement).toBeCalledWith('afterbegin', base) // for some reason the insertAdjacentElement doesn't work with jsdom. this test is a placeholder for when a better solution arises
+    })
+
+    test('should do nothing for absolute URLs', async () => {
+        const rootElement = window.document.createElement('div')
+        rootElement.innerHTML = '<a href="https://example.com/#home">Link</a>'
+        await fixLinks({rootElement, docUrl})
+        expect(rootElement.querySelector('*[href]').getAttribute('href')).toBe('https://example.com/#home')
+    })
+
+    test('should correct relative URLs', async () => {
+        const rootElement = window.document.createElement('div')
+        rootElement.innerHTML = '<a href="#home">Link</a>'
+        await fixLinks({rootElement, docUrl})
+        expect(rootElement.querySelector('*[href]').getAttribute('href')).toBe('https://example.com/page#home')
+    })
+
+    test('should not alter inline javascript in href attribute', async () => {
+        const rootElement = window.document.createElement('div')
+        rootElement.innerHTML = `<a href="javascript:alert('Hello');">Link</a>`
+        await fixLinks({rootElement, docUrl})
+        expect(rootElement.querySelector('*[href]').getAttribute('href')).toBe(`javascript:alert('Hello');`)
+    })
+
+    test('should not alter mailto scripts in href attribute', async () => {
+        const rootElement = window.document.createElement('div')
+        rootElement.innerHTML = `<a href="mailto:someone@example.com">Link</a>`
+        await fixLinks({rootElement, docUrl})
+        expect(rootElement.querySelector('*[href]').getAttribute('href')).toBe(`mailto:someone@example.com`)
+    })
+
+    test('should not alter data urls in href attribute', async () => {
+        const datauri = 'data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAABGdBTUEAALGP C/xhBQAAAAlwSFlzAAALEwAACxMBAJqcGAAAAAd0SU1FB9YGARc5KB0XV+IA AAAddEVYdENvbW1lbnQAQ3JlYXRlZCB3aXRoIFRoZSBHSU1Q72QlbgAAAF1J REFUGNO9zL0NglAAxPEfdLTs4BZM4DIO4C7OwQg2JoQ9LE1exdlYvBBeZ7jq ch9//q1uH4TLzw4d6+ErXMMcXuHWxId3KOETnnXXV6MJpcq2MLaI97CER3N0 vr4MkhoXe0rZigAAAABJRU5ErkJggg=='
+        const rootElement = window.document.createElement('div')
+        rootElement.innerHTML = `<a href="${datauri}">Link</a>`
+        await fixLinks({rootElement, docUrl})
+        expect(rootElement.querySelector('*[href]').getAttribute('href')).toBe(datauri)
+    })
+})
--- a/src/freeze-dry/index.test.js
+++ b/src/freeze-dry/index.test.js
@@ -1,0 +1,37 @@
+/* eslint-env jest */
+/* eslint import/namespace: "off" */
+
+import freezeDry from 'src/freeze-dry/index'
+import * as inlineStyles from 'src/freeze-dry/inline-styles'
+import * as removeScripts from 'src/freeze-dry/remove-scripts'
+import * as inlineImages from 'src/freeze-dry/inline-images'
+import * as setContentSecurityPolicy from 'src/freeze-dry/set-content-security-policy'
+import * as fixLinks from 'src/freeze-dry/fix-links'
+import * as removeNoscripts from './remove-noscripts'
+
+beforeAll(() => {
+    inlineStyles.default = jest.fn()
+    removeScripts.default = jest.fn()
+    inlineImages.default = jest.fn()
+    fixLinks.default = jest.fn()
+    setContentSecurityPolicy.default = jest.fn()
+    removeNoscripts.default = jest.fn()
+})
+
+describe('freezeDry', () => {
+    test('should call the whenAllSettled with all the jobs', async () => {
+        expect.assertions(6)
+        await freezeDry()
+        expect(inlineStyles.default).toHaveBeenCalled()
+        expect(removeScripts.default).toHaveBeenCalled()
+        expect(inlineImages.default).toHaveBeenCalled()
+        expect(fixLinks.default).toHaveBeenCalled()
+        expect(setContentSecurityPolicy.default).toHaveBeenCalled()
+        expect(removeNoscripts.default).toHaveBeenCalled()
+    })
+
+    test('should return the HTML document as a string', async () => {
+        const html = await freezeDry()
+        expect(html).toBe('<html><head></head><body></body></html>')
+    })
+})
--- a/src/freeze-dry/inline-images.test.js
+++ b/src/freeze-dry/inline-images.test.js
@@ -1,0 +1,26 @@
+/* eslint-env jest */
+/* eslint import/namespace: "off" */
+
+import * as inlineImages from 'src/freeze-dry/inline-images'
+import * as common from 'src/freeze-dry/common'
+
+describe('getUrlsFromSrcset', () => {
+    test('should return URL for srcset', () => {
+        expect.assertions(3)
+        const srcset = 'https://example.com/background1.jpg 1x, https://example.com/background2.jpg 2x, https://example.com/background3.jpg 3x'
+        const urls = inlineImages.getUrlsFromSrcset(srcset)
+        expect(urls[0]).toBe('https://example.com/background1.jpg')
+        expect(urls[1]).toBe('https://example.com/background2.jpg')
+        expect(urls[2]).toBe('https://example.com/background3.jpg')
+    })
+})
+
+describe('inlineImages', () => {
+    test('should call inlineUrlsInAttributes', async () => {
+        common.inlineUrlsInAttributes = jest.fn()
+        const rootElement = window.document.implementation.createHTMLDocument()
+        const docUrl = 'about:blank'
+        await inlineImages.default({rootElement, docUrl})
+        expect(common.inlineUrlsInAttributes).toHaveBeenCalledTimes(3)
+    })
+})
--- a/src/freeze-dry/inline-styles.test.js
+++ b/src/freeze-dry/inline-styles.test.js
@@ -1,0 +1,60 @@
+/* eslint-env jest */
+
+import inlineStyles from 'src/freeze-dry/inline-styles'
+import * as common from 'src/freeze-dry/common'
+import { dataURLToBlob } from 'blob-util'
+
+const imageDataUri = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAAAAAA6fptVAAAACklEQVR4nGNiAAAABgADNjd8qAAAAABJRU5ErkJggg=='
+
+describe('inlineStyles', () => {
+    const parser = new DOMParser()
+    let imageBlob
+
+    beforeAll(async () => {
+        imageBlob = await dataURLToBlob(imageDataUri)
+    })
+
+    test('should return <style> tag with the fetched stylesheet', async () => {
+        const styleSheet = 'div{background-image: url("public/image/background.jpeg");}'
+        fetch.mockResponseOnce(new Blob([styleSheet]))
+        fetch.mockResponseOnce(imageBlob)
+        const doc = parser.parseFromString(
+            '<html><head><link rel="stylesheet" type="text/css" href="https://example.com/theme.css"></head></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
+        await inlineStyles({rootElement: doc.documentElement, docUrl})
+        expect(doc.querySelector('style').innerHTML).toBe(`div{background-image: url(${imageDataUri});}`)
+    })
+
+    test('should convert the url in <style> to dataUris', async () => {
+        expect.assertions(2)
+        const styleSheet = 'div{background-image: url("public/image/background.jpeg");}'
+        fetch.mockResponseOnce(new Blob([styleSheet]))
+        const styleSheetAsDataUri = `data:text/plain;charset=utf-8;base64,aHR0cHM6Ly9leGFtcGxlLmNvbS9wdWJsaWMvaW1hZ2UvYmFja2dyb3VuZC5qcGVn`
+        const spy = jest.spyOn(common, 'urlToDataUri').mockReturnValue(styleSheetAsDataUri)
+        const doc = parser.parseFromString(
+            '<html><head><style type="text/css">div{background-image: url("public/image/background.jpeg");}</style></head><div></div></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
+        await inlineStyles({rootElement: doc.documentElement, docUrl})
+        expect(spy).toHaveBeenCalled()
+        expect(doc.querySelector('style').innerHTML).toBe(`div{background-image: url(${styleSheetAsDataUri});}`)
+        spy.mockRestore()
+    })
+
+    test('should convert the url in "style" to dataUris', async () => {
+        expect.assertions(2)
+        const spy = jest.spyOn(common, 'urlToDataUri').mockReturnValue(imageDataUri)
+        const doc = parser.parseFromString(
+            '<html><div style="background-image: url(\'public/image/background.jpeg\');"></div></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
+        await inlineStyles({rootElement: doc.documentElement, docUrl})
+        expect(spy).toHaveBeenCalled()
+        expect(doc.querySelector('div').getAttribute('style')).toBe(`background-image: url(${imageDataUri});`)
+        spy.mockRestore()
+    })
+})
--- a/src/freeze-dry/remove-scripts.test.js
+++ b/src/freeze-dry/remove-scripts.test.js
@@ -1,0 +1,27 @@
+/* eslint-env jest */
+
+import removeScripts from 'src/freeze-dry/remove-scripts'
+
+describe('removeScripts', () => {
+    test('should remove script tags from the document', async () => {
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><head><script></script></head></html>',
+            'text/html'
+        )
+        const docUrl = 'about:blank'
+        await removeScripts({rootElement, docUrl})
+        expect(rootElement.getElementsByTagName('script').length).toBe(0)
+    })
+
+    test('should remove "on" handlers', async () => {
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><div onmouseover="handler()" onClick="handler()"></div></html>',
+            'text/html'
+        )
+        const docUrl = 'about:blank'
+        await removeScripts({rootElement, docUrl})
+        expect(rootElement.querySelector('div').attributes.length).toBe(0)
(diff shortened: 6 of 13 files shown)


reficul31:
The changes asked by @Sanjo have been impelmented. Although I am not sure of mock returning a small image in the data uri as I am currently very hazy on the implementation. 
It would be really helpful if we could work on the failing test as currently I have no idea why its failing. If any other changes are required feel free to review.

Treora:
I tried running the tests but I get another error than you or Travis get:

```
  ● Test suite failed to run

    SyntaxError: Unexpected token [
      
      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/ScriptTransformer.js:289:17)
      at Object.<anonymous> (setupJest.js:1:118)
```

Apparently it does not like >ES6 syntax in `jest-fetch-mock`, though it does not mind it in the tests themselves (which I suppose do get transpiled while dependencies do not). Any idea how this could go wrong for me (also after a fresh `npm install`) but not on e.g. Travis?

Treora:
As for the error with adding the `base` tag, the travis log reports this error:
`TypeError: head.insertAdjacentElement is not a function`

I guess this could be an underlying bug in the jsdom.. (update: I now see `insertAdjacentElement` is not very standardised yet (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/insertAdjacentElement#Specification)), so it would not be surprising if it is not implemented in jsdom at all)

reficul31:
@Treora I looked into jsdom implementation of the ```insertAdjacentElement```, it seems it is not yet implemented. I guess a modification to the test to by mocking the function call might solve the problem? Also, still looking into your other error, I can't seem to figure out why it is happening at your end, also it is kinda hard to replicate the error since I haven't encountered it myself. But I'll still keep looking into it.

Treora:
I just filed an [issue](https://github.com/tmpvar/jsdom/issues/1890) at jsdom. In the meantime, we could either skip the test or mock the call and check if it gets the expected arguments.

Treora:
@reficul31: I also found my problem with syntax errors: I used node 5, which did not eat the es6/7(?) syntax used by jest-fetch-mock. I don't think requiring people to use node >=6 is a problem.

reficul31:
@Treora Thanks for looking into the error, so do you think using jest-fetch-mock is worth the trouble or not? I'll modify the tests according to whether we are able to use fetch-mock or not.

ghost:
I think requiring node.js >= 6 is not a problem. We should document that in the readme "Running tests" section (doesn't exist yet). Fixing that would also be fairly easy (just add babel transpilation to the package).

ghost:
Regarding the `Element.insertAdjacentElement` implementation: [`Element.insertAdjacentHTML`](https://github.com/tmpvar/jsdom/blob/master/lib/jsdom/living/nodes/Element-impl.js#L359-L407) is already implemented in jsdom. The `Element.insertAdjacentElement` implementation is very similar.
So we can implement it fairly easily. Side node: Jest currently uses jsdom 9.x.x. The newest release is 11.x.x which does only support node.js >= 6.

reficul31:
@Sanjo I can't seem to make the test work. So, I took to mocking the function call. I'll keep working out a way to work on the test so that we can shift our focus from that test and work on other test's implementation.

src/freeze-dry/index.test.js
```diff
@@ -0,0 +1,38 @@
...
+        const rootElement = window.document.implementation.createHTMLDocument()
+        const docUrl = 'about:blank'
+        const jobs = [
+            removeScripts.default({rootElement}),
+            inlineStyles.default({rootElement, docUrl}),
+            inlineImages.default({rootElement, docUrl}),
+            fixLinks.default({rootElement, docUrl}),
+        ]
+        await freezeDry()
+        expect(spy).toHaveBeenCalledWith(jobs)
```
ghost: This does not check if the four functions: removeScripts, inlineStyles, inlineImages, fixLinks were called from freezeDry because `jobs` is an array with `[undefined, undefined, undefined, undefined]`.

I think it should check that removeScripts, inlineStyles, inlineImages, fixLinks were called from freezeDry.

src/freeze-dry/index.test.js
```diff
@@ -0,0 +1,38 @@
...
+import * as whenAllSettled from 'src/util/when-all-settled'
+import * as inlineStyles from 'src/freeze-dry/inline-styles'
+import * as removeScripts from 'src/freeze-dry/remove-scripts'
+import * as inlineImages from 'src/freeze-dry/inline-images'
+import * as fixLinks from 'src/freeze-dry/fix-links'
+
+beforeAll(() => {
+    inlineStyles.default = jest.fn()
+    removeScripts.default = jest.fn()
+    inlineImages.defualt = jest.fn()
```
ghost: Typo: `defualt`

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
...
+            throw new Error('mock error')
+        })
+        const dataUri = await urlToDataUri('example.com')
+        expect(spy).toHaveBeenCalled()
+        expect(dataUri).toBe('about:blank')
+        spy.mockRestore()
+    })
+})
+
+describe('inlineUrlsInAttributes', () => {
```
ghost: I think the inlined functions in `inlineUrlsInAttributes` should be extracted and tested separately.
This would remove tests for different elements that basically test the same and allows for more fine grained tests for the different parts.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
...
+        expect.assertions(2)
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="public/image/background.png" alt="background" /></body></html>',
+             'text/html'
+             )
+        const docUrl = 'https://example.com'
+        await inlineUrlsInAttributes({elements: 'img', attributes: 'src', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('img').getAttribute('data-original-src')).toBe('public/image/background.png')
+        expect(rootElement.querySelector('img').getAttribute('src')).toBe('data:;base64,')
```
ghost: Testing if actual data is put in would be great.

src/freeze-dry/fix-links.test.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-env jest */
+
+import fixLinks from 'src/freeze-dry/fix-links'
+
+describe('fixLinks', () => {
```
ghost, line 5: I think the two condition branches and the inlined "replaceUrl" function can be extracted and tested separately.

src/freeze-dry/fix-links.test.js
```diff
@@ -0,0 +1,31 @@
...
+        const rootElement = window.document.implementation.createHTMLDocument()
+        rootElement.querySelector('head').insertAdjacentElement = jest.fn()
+        const docUrl = 'about:blank'
+        await fixLinks({rootElement, docUrl})
+        const base = document.createElement('base')
+        base.href = docUrl
+        expect(rootElement.querySelector('head').insertAdjacentElement).toBeCalledWith('afterbegin', base)
+    })
+
+    test('should do nothing for correct URLs', async () => {
```
ghost: I would say "absolute" instead of "correct" because "correct" is vague.

src/freeze-dry/fix-links.test.js
```diff
@@ -0,0 +1,31 @@
...
+        const rootElement = window.document.createElement('div')
+        rootElement.innerHTML = '<a href="https://example.com/#home">Link</a>'
+        const docUrl = 'https://example.com/'
+        await fixLinks({rootElement, docUrl})
+        expect(rootElement.querySelector('*[href]').getAttribute('href')).toBe('https://example.com/#home')
+    })
+
+    test('should correct dirty URLs', async () => {
```
ghost: I would say "relative" instead of "dirty" because "dirty" is vague.

src/freeze-dry/fix-links.test.js
ghost, line 5: Maybe add some tests for the values:

* `javascript:alert('Hello');`
* `mailto:someone@example.com`
* `data:image/png;base64,
iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAABGdBTUEAALGP
C/xhBQAAAAlwSFlzAAALEwAACxMBAJqcGAAAAAd0SU1FB9YGARc5KB0XV+IA
AAAddEVYdENvbW1lbnQAQ3JlYXRlZCB3aXRoIFRoZSBHSU1Q72QlbgAAAF1J
REFUGNO9zL0NglAAxPEfdLTs4BZM4DIO4C7OwQg2JoQ9LE1exdlYvBBeZ7jq
ch9//q1uH4TLzw4d6+ErXMMcXuHWxId3KOETnnXXV6MJpcq2MLaI97CER3N0
vr4MkhoXe0rZigAAAABJRU5ErkJggg==`

and make sure that those are not altered.

src/freeze-dry/inline-images.test.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-env jest */
```
ghost, line 1: Add tests for getUrlsFromSrcset. (You need to export it for that.)

src/freeze-dry/inline-images.test.js
```diff
@@ -0,0 +1,26 @@
...
+        const spy = jest.spyOn(whenAllSettled, 'default')
+        const rootElement = window.document.implementation.createHTMLDocument()
+        const docUrl = 'about:blank'
+        const jobs = [
+            common.inlineUrlsInAttributes({...{}, rootElement, docUrl}),
+            common.inlineUrlsInAttributes({...{}, rootElement, docUrl}),
+            common.inlineUrlsInAttributes({...{}, rootElement, docUrl}),
+        ]
+        await inlineImages({rootElement, docUrl})
+        expect(spy).toHaveBeenCalledWith(jobs)
```
ghost: I think this doesn't really test anything. I would rather test if the rootElement has been modified correctly by comparing the result with the expected DOM tree.

ghost (changes requested):

src/freeze-dry/inline-images.test.js
ghost: __Shivangs question:__

> 3:53] 
Your review at this line saying this doesn't test anything and you would rather test if the rootElement be modified correctly

> [3:54] 
I just had a small doubt, since these are unit tests, isn't it the gist of the unit test to make sure the code is executed in order without using external resources. If I made it such that I check for the modification of the rootElement then that would classify as an integrations test

> [3:55] 
Also it would require me to unmock the inlineUrlsInAttributes function as well

> [3:57] 
I'm not too sure of this, but in one of the best practices it was stated a unit tests should run comletely independent from rest of the code. Such that any change to external code shouldn't affect that particular test until changes are made to the module on which it is tested. In this case, if we change the inlineUrlsInAtrribute function, we will have to change the unit tests for this module as well

__My answer:__

Ok, so inlineUrlsInAttributes is tested in common.test.js which is fine. So forget about the " I would rather test if the rootElement has been modified correctly by comparing the result with the expected DOM tree.".

About the "I think this doesn't really test anything." I would suggest to expect that common.inlineUrlsInAttributes is called correctly.

src/freeze-dry/common.test.js
reficul31: I am sorry, but could you elaborate on that review a bit, I didn't really get the whole thing.

reficul31:
@Treora @Sanjo Any other changes required?

Treora:
@reficul31: Sorry I have not gotten around to reviewing your tests; either these or #109. I am not experienced with writing tests either, so I would have to sit down, look at the code itself again, and think what would actually be relevant scenarios to cover. If any of the mentors have thought about this too, I would be glad to hear their conclusions.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+fetch.mockResponse(new Blob([ {example: 'dummy'} ]), { 'status': 200, 'statusText': 'Response Accepted' })
+
+describe('urlToDataUri', () => {
+    test('should return a valid dataUri for a URL', async () => {
+        const spy = jest.spyOn(responseToDataUri, 'default')
+        await urlToDataUri('https://example.com')
+        expect(spy).toHaveBeenCalled()
+        spy.mockRestore()
+    })
```
Treora: This test does not look at the response at all. I'd have the `fetch` mock return some string, and test if the returned data URL corresponds to that.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
+/* eslint-env jest */
+
+import { inlineUrlsInAttributes, urlToDataUri } from 'src/freeze-dry/common'
+import * as responseToDataUri from 'src/util/response-to-data-uri'
+
+fetch.mockResponse(new Blob([ {example: 'dummy'} ]), { 'status': 200, 'statusText': 'Response Accepted' })
```
Treora: I would consider to instead use `fetch.mockResponseOnce` inside the tests themselves, to make behaviour of the test more obvious when reading a single test.
Is status 200 not the default by the way? Might be able to leave the second argument away.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
...
+        expect.assertions(2)
+        const spy = jest.spyOn(responseToDataUri, 'default').mockImplementation(() => {
+            throw new Error('mock error')
+        })
+        const dataUri = await urlToDataUri('example.com')
+        expect(spy).toHaveBeenCalled()
+        expect(dataUri).toBe('about:blank')
+        spy.mockRestore()
+    })
+})
```
Treora: In this test, I would probably let the `fetch` mock throw an error or return an error (4xx) response. In fact, we may want two separate tests for these. (besides, in the tests for `util/responseToDataUri` we should verify that an error response makes it throw an exception)

I think it should actually result in `about:invalid` rather than `about:blank` (concluding this from their [descriptions](https://www.iana.org/assignments/about-uri-tokens/about-uri-tokens.xhtml)) ; we should change the code accordingly.

src/freeze-dry/common.test.js
Treora: Also, we might want to add a test: given a `data:` URL, it should return that same URL.

src/freeze-dry/common.test.js
Treora: Sanjo has a point here: the mapping function ([here](https://github.com/WebMemex/webmemex-extension/blob/91d1b867c37398cfefd549dca8191534531282d8/src/freeze-dry/common.js#L33)) could be extracted and given a name, then tested by itself. However we would need to add it to exports, or find a way to access it from the test. (something like [rewire](https://stackoverflow.com/questions/14874208/how-to-access-and-test-an-internal-non-exports-function-in-a-node-js-module#30794280), but that does not support babel..)

I don't think your current approach is too bad either, though it does not test combinations of tags and attributes. I may however be to blame for creating such a function with many options; we could just keep these tests simple for now and refactor the code later to make it easier to test.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
...
+
+describe('inlineUrlsInAttributes', () => {
+    test('should change the URL in <img> tag to a dataUri', async () => {
+        expect.assertions(2)
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="public/image/background.png" alt="background" /></body></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
```
Treora: docUrl could be factored out, one level up, to the `describe` block, so you don't have to repeat it every time. Maybe likewise for `parser`, if it does not need to be initialised again each time (or you could define a small wrapper `htmlToDom` to that calls `parser.parseFromString` for you).

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,78 @@
...
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="background.png" alt="background" srcset="background.png 200w,background.png 400w"></body></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
+        await inlineUrlsInAttributes({elements: 'img', attributes: 'srcset', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('img').getAttribute('data-original-srcset')).toBe('background.png 200w,background.png 400w')
+        expect(rootElement.querySelector('img').getAttribute('srcset')).toBe('data:;base64,W29iamVjdCBPYmplY3Rd')
+    })
```
Treora: For `srcset`, we would want to end up with multiple data URLs (two in this case), for which we should pass a function to `attrsToUrl`, like in [inlineImages](https://github.com/WebMemex/webmemex-extension/blob/91d1b867c37398cfefd549dca8191534531282d8/src/freeze-dry/inline-images.js#L21). I don't know if we want to test it with this level of detail here though, perhaps that should be done in inline-images instead. In this test, it woul be fine to just test whether the `attrToUrls` function is used correctly (e.g. pass `attrToUrls: value => 'test_' + value` to `inlineUrlsInAttributes`).

src/freeze-dry/fix-links.test.js
```diff
@@ -0,0 +1,56 @@
...
+describe('fixLinks', () => {
+    test('should call the function to prepend the base element', async () => {
+        const rootElement = window.document.implementation.createHTMLDocument()
+        rootElement.querySelector('head').insertAdjacentElement = jest.fn()
+        const docUrl = 'about:blank'
+        await fixLinks({rootElement, docUrl})
+        const base = document.createElement('base')
+        base.href = docUrl
+        expect(rootElement.querySelector('head').insertAdjacentElement).toBeCalledWith('afterbegin', base)
+    })
```
Treora: Maybe add a comment explaining why the funny approach (jsdom's lack of `insertAdjacentElement`).

src/freeze-dry/index.test.js
```diff
@@ -0,0 +1,31 @@
...
+
+describe('freezeDry', () => {
+    test('should call the whenAllSettled with all the jobs', async () => {
+        expect.assertions(4)
+        await freezeDry()
+        expect(inlineStyles.default).toHaveBeenCalled()
+        expect(removeScripts.default).toHaveBeenCalled()
+        expect(inlineImages.default).toHaveBeenCalled()
+        expect(fixLinks.default).toHaveBeenCalled()
+    })
```
Treora: This test does not look useful to me as it is. I think it may be helpful to instead have a simple example document to test with, not as a test that covers all aspects but just as a sanity check; it could be just two lines of html. E.g.
```
inputDoc = parser.parseString('<html><head><script src="script.js"></script></head><body><img src="img.png" /></body></html>')
expectedDoc = '<html><head><base href="http://example.org" /></head><body><img src="data:image/png,base64,ab1234" original-src="img.png" /></body></html>'
```
And then test if `await freezeDry(inputDoc, 'http://example.org') === expectedDoc`. Or something like that. I do realise that this the output would depend on the inner functions, and would thus have to be updated rather frequently, so tell me if you think it's not a good idea.

Also we could test that the original DOM is not modified.

src/freeze-dry/inline-images.test.js
```diff
@@ -0,0 +1,24 @@
...
+import * as common from 'src/freeze-dry/common'
+
+describe('getUrlsFromSrcset', () => {
+    test('should return URL for srcset', () => {
+        expect.assertions(3)
+        const srcset = 'https://example.com/background.jpg 1x, https://example.com/background.jpg 2x, https://example.com/background.jpg 3x'
+        const urls = inlineImages.getUrlsFromSrcset(srcset)
+        urls.map(url => expect(url).toBe('https://example.com/background.jpg'))
+    })
+})
```
Treora: Nice you added a test for this; I would however take several different URLs, not have them all be the same.

src/freeze-dry/inline-images.test.js
```diff
@@ -0,0 +1,24 @@
...
+
+describe('inlineImages', () => {
+    test('should call inlineUrlsInAttributes', async () => {
+        common.inlineUrlsInAttributes = jest.fn()
+        const rootElement = window.document.implementation.createHTMLDocument()
+        const docUrl = 'about:blank'
+        await inlineImages.default({rootElement, docUrl})
+        expect(common.inlineUrlsInAttributes).toHaveBeenCalledTimes(3)
+    })
+})
```
Treora, line 26: Seems rather arbitrary to just test whether something was called three times. Like with `index.js`, I would consider giving some example doc(s) and checking the results.

src/freeze-dry/inline-styles.js
```diff
@@ -44,7 +44,7 @@ async function inlineLinkedStylesheets({rootElement, docUrl}) {
         } catch (err) {
             stylesheetText = '/* Oops! Freeze-dry failed to save this stylesheet. */'
         }
-        const styleEl = rootElement.ownerDocument.createElement('style')
+        const styleEl = rootElement.createElement('style')
```
Treora: Why this change? `rootElement` should be an element, not a document. Perhaps check your tests if you are passing a document to it anywhere, and pass `document.documentElement` as rootElement instead.

src/freeze-dry/inline-styles.test.js
```diff
@@ -0,0 +1,51 @@
+/* eslint-env jest */
+
+import inlineStyles from 'src/freeze-dry/inline-styles'
+import * as common from 'src/freeze-dry/common'
+
+fetch.mockResponse(new Blob(['Dummy Styles']), { 'status': 200, 'statusText': 'Response Accepted' })
+
+describe('inlineStyles', () => {
+    test('should return create a <style> with the fetch of stylesheet', async () => {
```
Treora: broken sentence.

src/freeze-dry/inline-styles.test.js
```diff
@@ -0,0 +1,51 @@
...
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><head><link rel="stylesheet" type="text/css" href="https://example.com/theme.css"></head></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
+        await inlineStyles({rootElement, docUrl})
+        expect(rootElement.querySelector('style').innerHTML).toBe('Dummy Styles')
+    })
```
Treora: Great, this is the kind of test I mean with passing an example doc and checking the result.

src/freeze-dry/inline-styles.test.js
```diff
@@ -0,0 +1,51 @@
...
+        const rootElement = parser.parseFromString(
+            '<html><head><style type="text/css">div{background-image: url("public/image/background.jpeg");}</style></head><div></div></html>',
+            'text/html'
+        )
+        const docUrl = 'https://example.com'
+        await inlineStyles({rootElement, docUrl})
+        expect(spy).toHaveBeenCalled()
+        expect(rootElement.querySelector('style').innerHTML).toBe(`div{background-image: url(${dataUri});}`)
+        spy.mockRestore()
+    })
```
Treora: Great. I would perhaps just extract the long strings, and make it obvious to the reader that the data URL matches the stylesheet contents; e.g.:
```
const styleSheet = 'div{background-image: url("public/image/background.jpeg");}'
const styleSheetAsDataUri = `data:text/plain;charset=utf-8;base64,aHR0cHM6Ly9leGFtcGxlLmNvbS9wdWJsaWMvaW1hZ2UvYmFja2dyb3VuZC5qcGVn`
```
(or use [`btoa`](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/btoa) to convert the string on the spot, but note it has trouble with unicode, string must remain ascii)

Treora (commented):
Good stuff! Some stuff to be changed, perhaps have a read through my comments and a look at the tests again.

Small note; I would everywhere make docUrl e.g. `https://example.com/page`, not just `https://example.com`, so that we would e.g. detect if we mix it up with the base URL.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,85 @@
...
+        spy.mockRestore()
+    })
+
+    test('should return a "about:invalid" for an error URL' ,async () => {
+        fetch.mockResponseOnce(new Error())
+        const dataUri = await urlToDataUri('example.com')
+        expect(dataUri).toBe('about:blank')
+    })
+
+    test('should return the same url if already in dataUri form', async () => {
```
reficul31: According to @Treora review I added the two tests. I hope these were the test cases asked. But for some reason the tests seem to both fail. Any input would be helpful

src/freeze-dry/index.test.js
reficul31: @Treora These tests are called unit tests. The nature of the tests are such that they are not supposed to use an external resources and just check if the code traces the steps as it is supposed. The tests are designed in such a manner because of the no external dependency policy, which means if a change in the code takes place it should only effect the test that runs on it.
I think what you are referring to are integrations tests, the testing team will be working on them next and they will be soon implemented.

src/freeze-dry/inline-images.test.js
reficul31: These tests are necessary to check whether the code runs the way we want it to run. I know it is a bit restrictive, but I wrote the tests on the grounds of unit testing.

src/freeze-dry/inline-styles.js
reficul31: I kept encountering an error with this exact statement, which I did not before it was changed. So I started fiddling with the code to hack my way through, but no change in the testing code made it run. If there are any suggestions, I will be extremely happy to hear them.

reficul31 (commented):
Thanks for the review @Treora and @Sanjo, I had some questions regarding the code which I have put under this review. Feel free to suggest any other changes that are required.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+    test('should return a "about:invalid" for an error URL', async () => {
+        fetch.mockResponseOnce(new Error())
+        const dataUri = await urlToDataUri('example.com')
+        expect(dataUri).toBe('about:invalid')
+    })
+
+    test('should return the same url if already in dataUri form', async () => {
+        const dataUri = await urlToDataUri('data:;base64,test')
+        expect(dataUri).toBe('data:;base64,')
```
Treora: The response should be equal to the input, so including the word `test` here. This test would work in the browser, with the real fetch implementation, but this won't work while fetch is a mock. We could let the fetch mock return a known string, but then it would not test much really. I guess I wanted to test the behaviour of `fetch` itself, which is out of scope. Perhaps we should remove this test after all?

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+            throw new Error('mock error')
+        })
+        const dataUri = await urlToDataUri('example.com')
+        expect(spy).toHaveBeenCalled()
+        expect(dataUri).toBe('about:invalid')
+        spy.mockRestore()
+    })
+
+    test('should return a "about:invalid" for an error URL', async () => {
+        fetch.mockResponseOnce(new Error())
```
Treora: Look like a weird and unpredictable way to may fetch throw an error (does it even?). Is `mockReject` perhaps a better solution? (scarcely documented though, but used in this [example](https://github.com/jefflau/jest-fetch-mock#example-2---mocking-a-failed-fetch))

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+        const dataUri = await urlToDataUri('example.com')
```
Treora: Put http(s) before the URL (here and above), otherwise it serves as a wrong example of usage.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+        const spy = jest.spyOn(responseToDataUri, 'default').mockImplementation(() => {
+            throw new Error('mock error')
+        })
+        const dataUri = await urlToDataUri('example.com')
+        expect(spy).toHaveBeenCalled()
+        expect(dataUri).toBe('about:invalid')
+        spy.mockRestore()
+    })
+
+    test('should return a "about:invalid" for an error URL', async () => {
```
Treora: 'for an error URL'  => 'when fetching fails'.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+
+describe('urlToDataUri', () => {
+    test('should return a valid dataUri for a URL', async () => {
+        const spy = jest.spyOn(responseToDataUri, 'default')
+        await urlToDataUri('https://example.com/page')
+        expect(spy).toHaveBeenCalled()
+        spy.mockRestore()
+    })
+
+    test('should return a "about:invalid" for a invalid URL', async () => {
```
Treora: 'invalid URL'? Perhaps just 'upon failure' or something? This tests what happens when `responseToDataUri` throws, right? Not sure when that would happen, but it's not necessarily the mistake of the URL.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+        expect(dataUri).toBe('data:;base64,')
+    })
+})
+
+describe('inlineUrlsInAttributes', () => {
+    const docUrl = 'https://example.com/page'
+    const parser = new DOMParser()
+
+    test('should change the URL in <img> tag to a dataUri', async () => {
+        fetch.mockResponseOnce(new Blob([ {example: 'Lorem Ipsum'} ]))
```
Treora: A bit confusing to use text here. Please make an actual tiny image blob (could be a global object for using in more tests), with mime type, and a constant with its data url. See e.g. [these lines](https://github.com/nolanlawson/blob-util/blob/52364427b15ba6d82eb77d28f516d0d44a71be3e/test/test.js#L16-L18).

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+        expect.assertions(2)
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="public/image/background.png" alt="background" /></body></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'img', attributes: 'src', fixIntegrity: true, rootElement, docUrl})
```
Treora: I would leave irrelevant arguments (`fixIntegrity` here) away, to not distract attention.

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+        fetch.mockResponseOnce(new Blob([ {example: 'Lorem Ipsum'} ]))
+        expect.assertions(2)
+        const rootElement = parser.parseFromString(
+            '<html><head><link rel="icon" href="public/image/favicon.ico"></head></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'link', attributes: 'href', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('link').getAttribute('data-original-href')).toBe('public/image/favicon.ico')
+        expect(rootElement.querySelector('link').getAttribute('href')).toBe('data:;base64,W29iamVjdCBPYmplY3Rd')
+    })
```
Treora: Is it worth adding this test? It looks like only the tag name is different (`img` → `link`). We could end up adding tests for every element type. ;)

src/freeze-dry/common.test.js
```diff
@@ -0,0 +1,97 @@
...
+        fetch.mockResponseOnce(new Blob([ {example: 'Lorem Ipsum'} ]))
+        expect.assertions(2)
+        const rootElement = parser.parseFromString(
+            '<html><body><img src="background.png" alt="background" srcset="background.png 200w,background.png 400w"></body></html>',
+            'text/html'
+        )
+        await inlineUrlsInAttributes({elements: 'img', attributes: 'srcset', fixIntegrity: true, rootElement, docUrl})
+        expect(rootElement.querySelector('img').getAttribute('data-original-srcset')).toBe('background.png 200w,background.png 400w')
+        expect(rootElement.querySelector('img').getAttribute('srcset')).toBe('data:;base64,W29iamVjdCBPYmplY3Rd')
+    })
```
Treora: Same as above: no apparent added value. We could add a test for the attrToUrls option instead, which is used for attributes such as `srcset` with multiple URIs in them.

src/freeze-dry/fix-links.test.js
Treora: Agreed.

src/freeze-dry/inline-styles.js
Treora: You have been mixing up documents and their `<html>` element (= document.documentElement). DOMParser returns a document, not the documentElement. Check and fix all assignments to rootElement and the problem should disappear.

src/freeze-dry/inline-styles.test.js
```diff
@@ -0,0 +1,53 @@
...
+import inlineStyles from 'src/freeze-dry/inline-styles'
+import * as common from 'src/freeze-dry/common'
+
+describe('inlineStyles', () => {
+    const parser = new DOMParser()
+
+    test('should return <style> tag with the fetched stylesheet', async () => {
+        const styleSheet = 'div{background-image: url("public/image/background.jpeg");}'
+        fetch.mockResponseOnce(new Blob([styleSheet]))
+        fetch.mockResponseOnce(new Blob(['background image']))
```
Treora: Use a real image for clarity.

src/freeze-dry/set-content-security-policy.test.js
```diff
@@ -0,0 +1,18 @@
...
+        const policyDirectives = [
+            "default-src 'none'",
+            "img-src data:",
+            "style-src data: 'unsafe-inline'",
+            "font-src data:",
+        ]
+        const doc = window.document.implementation.createHTMLDocument()
+        doc.querySelector('head').insertAdjacentElement = jest.fn()
+        setContentSecurityPolicy({doc, policyDirectives})
+        expect(doc.querySelector('head').insertAdjacentElement).toHaveBeenCalled()
```
Treora, line 16: I would rather check if the first child in the `head` is the meta tag with values as expected.

src/freeze-dry/remove-scripts.test.js
```diff
@@ -0,0 +1,27 @@
...
+        )
+        const docUrl = 'about:blank'
+        await removeScripts({rootElement, docUrl})
+        expect(rootElement.getElementsByTagName('script').length).toBe(0)
+    })
+
+    test('should remove "on" handlers', async () => {
+        const parser = new DOMParser()
+        const rootElement = parser.parseFromString(
+            '<html><div onHover="handler()" onClick="handler()"></div></html>',
```
Treora: Not a big problem, but `onHover` or `onhover` does not exist. `onmouseover` does. `onclick` is lowercase.

Treora (commented):
Another wave of comments. If you have looked over it all again and tackle some of these, perhaps we could open a hackpad or something and fix any remaining things together.

src/freeze-dry/set-content-security-policy.test.js
reficul31: If you have a look at the previous discussions, we established that this method was for some reason not working with jsdom. Therefore, I took to mocking the call to the function and checking if it is called.

Treora: Ah yes I forgot about that issue. Best to add a quick comment about it for that reason. Perhaps with a link to https://github.com/tmpvar/jsdom/issues/1890.

Treora:
I committed the changes we made in our call in the `test-freezedry` branch (commit 314f069). You could incorporate them in this PR, possibly still change things if you wish to, but if you think it's good then I'll merge.

PR #109, opened 2017-06-16
Unit tests for activity-logger module

reficul31:
Initial implementation of the activity logger tests.


reficul31:
@Sanjo @Treora 
Any inputs on the tests?

src/activity-logger/background/log-page-visit.test.js
```diff
@@ -0,0 +1,40 @@
...
+
+import maybeLogPageVisit from 'src/activity-logger/background/log-page-visit'
+import * as storePage from 'src/page-storage/store-page'
+import db from 'src/pouchdb'
+
+afterAll(async () => {
+    await db.destroy()
+})
+
+describe('maybeLogPageVisit', () => {
```
ghost, line 17: Maybe add a test for the call to `storeVisit`.
I would also suggest to extract a `logPageVisit` function from `maybeLogPageVisit`. E.g.:

```js
export async function maybeLogPageVisit({tabId, url}) {
    // Check if we want to log this page (hence the 'maybe' in the name).
    if (isWorthRemembering({url})) {
        await logPageVisit({tabId, url})
    }
}

export async function logPageVisit({tabId, url}) {
    // The time to put in documents.
    const timestamp = Date.now()

    // TODO first try to extend an existing visit instead of logging a new one.

    // First create an identifier for the page being visited.
    const {page, finalPagePromise} = await reidentifyOrStorePage({tabId, url})
    // Create a visit to this page.
    const visit = await storeVisit({page, url, timestamp})

    // TODO possibly deduplicate the visit if it was to the same page after all.
    void (finalPagePromise, visit)
}
```

Treora: Hah, I did exactly this refactoring last night (then to make a manual log-button). The code for the logging flow (including analysis/deduplication) could use more refactoring like this. Please feel free to (suggest to) refactor things while writing tests, as that seems a good moment to discover a more sensible structure.

reficul31: @Treora should I wait for the changes to be pushed and then proceed to rebase and test, or would you like me to refactor the code with the tests and modify the pull request?

reficul31: @Sanjo I left out the test for storeVisit as it was not included in the public API(it isn't exported). Do you think we should test the functions which are not exported? @Treora any opinions?

Treora: I would definitely like to test some functions that are not exported. The question is how to do this.
- One option is something like [rewire](https://github.com/jhnns/rewire), but it should work with our babel setup, which rewire does not support.
- Another option is to export everything (that we wish to test) from each file, but then have extra files that import and re-export only the ones that are really meant to be exported. Most logically the public API of files in a folder would be exposed in `index.js`, which could perhaps pull and export the default exports from files around it. It's just one logical structure that pops up in my mind now.
- A third option is to just export things you want to test and not bother about the difference of public/private functions for now. I suggest you take this approach now, and we move to a neater approach later (probably as part of a larger organisational restructuring, see #112).

Treora: As for the refactoring of this particular function, I will probably push it tomorrow or so, as I intend to use this in the release, so it might be easier for you to just rebase on that then.

ghost: I agree with the approach that @Treora suggests. Finding out who imports an exported function is quiet easy in the case of function signature changes.

reficul31: @Treora @Sanjo 
I have been reading up about rewire. You guys mind if I give it a try?

ghost: I would not recommend to use rewire because of the [limitations](https://github.com/jhnns/rewire#limitations). Also it is implemented for CommonJS which is not future proof for when ES modules are supported natively.

reficul31:
@Sanjo @Treora 
For now I have left the decision of testing the non exported functions, we could have a discussion in the future to decide the fate of non exported function testing.
I have update the PR with the tests asked for the logPageVisit function after the refactoring done by @Treora . Please review and do tell if any more changes in tests are required.

Treora:

PR #115, opened 2017-07-07
Unit tests for page-analysis module

reficul31:


Treora:
Closing this PR. Partly outdated, partly adding more maintenance work than value; I do like the regression test for the extracted page (meta)data, I just added that in commit 0ed9d17.

PR #107, opened 2017-06-09
[Work in Progress] Testing Suite

reficul31:
[Refs](https://github.com/WebMemex/webmemex-extension/issues/103)

**NOTE:** Do not merge. Work in progress for the testing suite built with jest.

## Status
- Activity Logger (Ready for review)
- Freeze Dry Module(Ready for Review)
- Options Module(Ready for Review)
- Page Analysis Module(Input Required)
- Util Module(Input Required)

PR #116, opened 2017-07-11
Unit tests for Util Module

reficul31:
There is definitely some room for improvement(least to say). I have tried implementing everything @Sanjo instructed [here](https://github.com/WebMemex/webmemex-extension/pull/107#issuecomment-313047062).


package.json
```diff
@@ -91,7 +91,9 @@
     "pify": "^3.0.0",
     "postcss-cssnext": "^2.11.0",
     "pouchdb-adapter-memory": "^6.2.0",
-    "pouchdb-node": "^6.2.0",
+    "pouchdb-core": "^6.3.4",
+    "pouchdb-mapreduce": "^6.3.4",
+    "pouchdb-replication": "^6.3.4",
```
Treora, line 96: great change, pouchdb-node pulls in way more dependencies than we need.

setupJest.js
```diff
@@ -1 +1,51 @@
...
+        case "beforeend": {
+            this.appendChild(fragment.content);
+            break;
+        }
+        case "afterend": {
+            this.parentNode.insertBefore(fragment.content, this.nextSibling);
+            break;
+        }
+    }
+}
```
Treora: We ended up with a cleaner version of this in `freeze-dry`, would be good to use that code here too.

src/util/delay.test.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-env jest */
+
+import delay from 'src/util/delay'
+jest.useFakeTimers()
```
Treora: Is it clean to do this here? Or should we do this in a `beforeAll` or something, end restore the situation afterwards? I guess we won't want real timers in our tests, so perhaps it does not matter. I'll leave you to decide.

src/util/delay.test.js
```diff
@@ -0,0 +1,13 @@
...
+
+describe('delay tests', () => {
+    test('should return with the given delay', () => {
+        expect.assertions(2)
+        delay(1000)
+        expect(setTimeout.mock.calls.length).toBe(1)
+        expect(setTimeout.mock.calls[0][1]).toBe(1000)
+    })
+})
```
Treora: This does not test whether delay resolves only after the given time. Looking at this [jest timer guide](https://facebook.github.io/jest/docs/timer-mocks.html) it seems that a better test should be possible. E.g. call `delay(1000)`, fast-forward time to 999, and assert it has not been resolved; then fast-forward the time to 1000 and assert that is has resolved.

src/util/event-to-promise.test.js
```diff
@@ -0,0 +1,37 @@
...
+        const resolveOpt = returnResolveOpts('test')
+        const rejectOpt = returnRejectOpts('Error')
+        eventToPromise({
+            resolve: resolveOpt,
+            reject: rejectOpt,
+        })
+        expect(resolveOpt.event.addListener).toBeCalled()
+        expect(rejectOpt.event.addListener).toBeCalled()
+    })
+})
```
Treora: This tests very little of the behaviour. That's okay to start with, but it may be good to add a comment (`// TODO `...) to not give the impression that this function is adequately tested.

src/util/delay.test.js
```diff
@@ -0,0 +1,13 @@
+/* eslint-env jest */
+
+import delay from 'src/util/delay'
+jest.useFakeTimers()
+
+describe('delay tests', () => {
```
Treora: nitpick: ` tests` can be removed from the description (also in the other files); the idea is that it makes a full sentence when combined with the `should ...` sentences below.

src/util/make-range-transform.test.js
```diff
@@ -0,0 +1,49 @@
...
+        expect.assertions(2)
+        const transformFunction1 = makeRangeTransform({
+            domain: [0, 100],
+            range: [0, 200],
+            clampOutput: false,
+        })
+        expect(transformFunction1(50)).toBe(100)
+        const transformFunction2 = makeRangeTransform({
+            domain: [0, 200],
+            range: [0, 100],
```
Treora: Please try some domain and range that do not start from `0`, and that start from a different number. Those are typical cases a faulty implementation could accidentally do right.

src/util/make-range-transform.test.js
```diff
@@ -0,0 +1,49 @@
...
+            clampOutput: false,
+        })
+        expect(transformFunction2(50)).toBe(25)
```
Treora: Please run the functions with a few different values (also in the tests below). least one below the domain, one within the domain, and one above the domain. Only then will we see the difference between the clamped and unclamped versions.

src/util/make-range-transform.test.js
```diff
@@ -0,0 +1,49 @@
...
+            clampOutput: true,
+        })
+        expect(transformFunction2(103)).toBeCloseTo(4.41, 2)
+    })
+})
+
+describe('makeNonlinearTransform tests', () => {
+    test('should return non linear transform function', () => {
+        const transformFunction = makeNonlinearTransform({
+            domain: [1000 * 60 * 5, 1000 * 60 * 60 * 24],
```
Treora: You can just choose any numbers here, no need to take ones from the source code.

src/util/make-range-transform.test.js
```diff
@@ -0,0 +1,49 @@
...
+            range: [0, 100],
+            clampOutput: true,
+            nonlinearity: Math.log,
+        })
+        expect(transformFunction(60 * 60 * 1000)).toBeCloseTo(43.88, 2)
+    })
+})
```
Treora: Perhaps we should (for either or both functions) add a test where the domain and/or range go from a high to a low number, e.g. `[100, 50]`. Seems a useful test since I am not even certain whether the current implementation will handle that properly!
We could also add negative number(s) somewhere.

src/util/nice-time.test.js
```diff
@@ -0,0 +1,55 @@
...
+        const now = new Date()
+        now.setHours(date.getHours() + 5)
+        expect(niceTime(date, {now})).toBe(`${('0' + date.getHours()).slice(-2)}:${('0' + date.getMinutes()).slice(-2)}`)
+    })
+
+    test('should return the timeperiod stamp and the day for timeperiod less than 24 hours but not on the same day', () => {
+        const date = new Date()
+        const now = new Date()
+        now.setHours(date.getHours() + 22)
+        expect(niceTime(date, {now})).toBe(`Yesterday ${('0' + date.getHours()).slice(-2)}:${('0' + date.getMinutes()).slice(-2)}`)
```
Treora: This test (and others) does too much computation (we should not have to test the test itself!). Just use a known date, e.g. `now = new Date(2016, 7, 2, 14, 55)`, and then expect a fixed string, e.g. `toBe('Yesterday 14:55')`.

src/util/event-to-promise.test.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-env jest */
+
+import eventToPromise from 'src/util/event-to-promise'
```
Treora: Since these files are next to each other, it seems easier to just use `./event-to-promise` instead. (and likewise in the other files)

src/util/pouchdb-update-doc.test.js
```diff
@@ -0,0 +1,17 @@
...
+describe('updateDoc', () => {
+    test('should call the database functions', async () => {
+        expect.assertions(2)
+        db.put = jest.fn()
+        db.get = jest.fn()
+        const docId = 'page/1234567890123/1234567890'
+        const updateFunc = jest.fn()
+        await updateDoc(db, docId, updateFunc)
+        expect(db.get).toHaveBeenCalledWith(docId)
+        expect(db.put).toHaveBeenCalledTimes(1)
```
Treora, line 14: Should we verify that `updateFunc` is called with the resolved return value from `db.get`, and `db.put` with the resolved return value from `updateFunc`?

src/util/short-url.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+    test('should remove "http" prefix if url is within maxLength', () => {
+        const url = shortUrl('https://example.com/page')
+        expect(url).toBe('example.com/page')
+    })
+
+    test('should do nothing to url with no prefix and within maxLength', () => {
+        const url = shortUrl('example.com/page')
+        expect(url).toBe('example.com/page')
+    })
```
Treora: I would remove this test because, even though it works, I would not recommend people to pass a URL without its scheme (which technically speaking is probably not really a URL any more).

src/util/short-url.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+    test('should slice the data uri', () => {
+        const dataUri = 'data:text/plain;charset=utf-8;base64,aHR0cHM6Ly9leGFtcGxlLmNvbS9wdWJsaWMvaW1hZ2UvYmFja2dyb3VuZC5qcGVn'
+        const url = shortUrl(dataUri)
+        expect(url).toBe(dataUri.slice(0, 47)+'...')
```
Treora: Replace the computation with the actual string, makes it more readable (and as commented on the `niceTime` tests, less computation reduces possibility of coding mistakes inside the test)

src/util/sync-location-hashes.test.js
```diff
@@ -0,0 +1,27 @@
...
+
+import syncLocationHashes from 'src/util/sync-location-hashes'
+
+function createWindowStub({hash}) {
+    return [{
+        location: {hash},
+        addEventListener: jest.fn(),
+        removeEventListener: jest.fn(),
+    }]
+}
```
Treora, line 11: I would create two window mocks here instead of one, so we can (later) add tests that test whether they properly sync to each other.

src/util/sync-location-hashes.test.js
```diff
@@ -0,0 +1,27 @@
...
+        disableListener()
+    })
+
+    test('should disable the listener when return function is called', () => {
+        const windows = createWindowStub({hash: 'home'})
+        const disableListener = syncLocationHashes(windows, {initial: undefined})
+        disableListener()
+        expect(windows[0].removeEventListener).toBeCalled()
+    })
+})
```
Treora: Maybe add a short `TODO` comment, telling there's much to still be tested.

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+        await whenPageDOMLoaded({tabId})
+        expect(browser.tabs.executeScript).toHaveBeenCalledWith(tabId, {code: 'undefined', runAt: 'document_end'})
+    })
+
+    test('should reject the promise if the script is not executed', async () => {
+        browser.tabs = {
+            executeScript: jest.fn().mockReturnValue(
+                new Promise((resolve, reject) => {
+                    reject(new Error('Script unable to execute'))
+                })
```
Treora: `Promise.reject('...')`

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+        }
+    })
+
+    test('should resolve promise when DOM page is loaded', async () => {
+        expect.assertions(2)
+        browser.tabs = {
+            executeScript: jest.fn().mockReturnValue(
+                new Promise((resolve, reject) => {
+                    resolve()
+                })
```
Treora: Just use `Promise.resolve()` (also below, and maybe elsewhere)

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+            ),
+        }
+        await whenPageDOMLoaded({tabId})
+        expect(browser.tabs.executeScript).toHaveBeenCalledWith(tabId, {code: 'undefined', runAt: 'document_end'})
+    })
+
+    test('should reject the promise if the script is not executed', async () => {
```
Treora, line 29: I think this test does actually need a `expect.assertions(1)`, since if the promise would mistakenly resolve, the assertion will be missed, and the test will pass (or does jest complain if a test has not performed any assertions?).
All other tests I have read so far seem to declare their number of assertions needlessly, since the assertions are in the main flow of the test; please remove those lines everywhere.

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+                    resolve()
+                })
+            ),
+        }
+        await whenPageDOMLoaded({tabId})
+        expect(browser.tabs.executeScript).toHaveBeenCalledWith(tabId, {code: 'undefined', runAt: 'document_end'})
+        expect(eventToPromise.default).toHaveBeenCalled()
+    })
+
+    test('should reject the promise if tab is changed', async () => {
```
Treora, line 54: `expect.assertions(1)`

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
+/* eslint-env jest */
+/* eslint import/namespace: "off" */
+
+import { whenPageDOMLoaded, whenPageLoadComplete, whenTabActive } from 'src/util/tab-events'
+import * as eventToPromise from 'src/util/event-to-promise'
+
+describe('whenPageDOMLoaded', () => {
+    const tabId = 1
+
+    beforeAll(() => {
```
Treora: Would be better to make this `beforeEach`, to ensure a clean mock every time (you change the return values in some tests). (also check other tests/files)

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+            new Promise((resolve, reject) => {
+                resolve()
+            })
+        )
+        browser.webNavigation = {
+            onCommitted: jest.fn(),
+        }
+    })
+
+    test('should return if the tab status is complete', async () => {
```
Treora: 'return directly' would be clearer than 'return'

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+
+    test('should return a promise which resolves', async () => {
+        browser.tabs = {
+            get: jest.fn().mockReturnValueOnce({
+                status: 'incomplete',
+            }),
+        }
+        await whenPageLoadComplete({tabId})
+        expect(eventToPromise.default).toHaveBeenCalled()
+    })
```
Treora: This test is a bit misleading; it only works because you mocked `eventToPromise` to resolve immediately, which is not obvious when reading just this test. I would prefer to set this mock's return value explicitly inside this test (and others that rely on it), rather than doing it once for every test.
(ideally we would use mock promise that we can manually resolve, as we also discussed [here](https://github.com/Treora/when-all-settled/pull/1#pullrequestreview-51621120), but that can be done later).

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+
+    test('should reject a promise if tab is changed', async () => {
```
Treora: `expect.assertions(1)`.. also in rejection tests below.

Treora: 'if tab is changed' is too generic, it could sound like e.g. one is activating another tab. Something like 'if tab has changed location' gets closer.

src/util/tab-events.test.js
```diff
@@ -0,0 +1,185 @@
...
+        )
+        try {
+            await whenPageLoadComplete({tabId})
+        } catch (err) {
+            expect(err.toString()).toBe('Error: Tab was changed')
+        }
+    })
+})
+
+describe('whenTabActive', () => {
```
Treora, line 124: Similar comments apply to these tests as to the tests above.

src/util/tree-walker.test.js
```diff
@@ -0,0 +1,64 @@
+/* eslint-env jest */
```
Treora, line 1: I'll review this file another time.

src/util/webextensionRPC.test.js
```diff
@@ -0,0 +1,90 @@
+/* eslint-env jest */
```
Treora, line 1: I'll review this file another time.

Treora (commented):
Sorry I did not look at your tests in a while. Nice work so far. Let's try finalise and merge some (or all) of these util tests. Utility functions seem most important to have tested to me, so that we can just rely on them and forget about their implementations.

Since this is a lot of code, please make new commits with changes instead of squashing, otherwise I'll lose track of things. If it's not too much effort, it would actually be much easier to just create separate PRs for each file you test, so we can merge them one by one as they are ready and keep an easy overview of what's left to be done.

src/util/delay.test.js
reficul31: The motive of the tests wasn't to test whether or not the promise is resolved, it was to check whether the specified timeout was called. The test can be changes, but I felt this implementation was better for this particular module.

src/util/nice-time.test.js
reficul31: I thought the test should test all the cases of niceTime module. This would provide as a useful use case as well as make the system very stable. If we go the other way and just check with the single input, we might not come to know if some other part of the program gets broken.

src/util/pouchdb-update-doc.test.js
reficul31: I am sorry, I couldn't understand this review.

src/util/sync-location-hashes.test.js
reficul31: I don't get what you mean by two window mocks? Could you please elaborate?

src/util/tab-events.test.js
reficul31: Good catch. The test it seems was falsely passing all this time. Although I can't seem to figure out why it doesn't work. Any suggestions?

reficul31:
@Treora 
> If it's not too much effort, it would actually be much easier to just create separate PRs for each file you test, so we can merge them one by one as they are ready and keep an easy overview of what's left to be done.

Its no effort at all. Do you want me to close all the un-reviewed pull requests and make individual ones?

Treora:
> Its no effort at all. Do you want me to close all the un-reviewed pull requests and make individual ones?

@reficul31: I was talking just about this PR with tests for things in `/src/util/`. Others can be considered after that; I try to just tackle one at a time.

src/util/delay.test.js
Treora: It could also be implemented as a separate test. I'm not sure if this test still adds value if a more complete test is added.

src/util/nice-time.test.js
Treora: The way it is, each test also tests only a single value. When using `Date.now()` they may however test a different value every time you run the tests. This is not great for reliability: tests that may sometimes fail and sometimes pass would drive me crazy. Better to have it be more deterministic, and use less computation. Tests could use different dates to test various cases, but the dates and expected results would be hard-coded.

src/util/pouchdb-update-doc.test.js
Treora: I wondered if we could have something like (pseudocode):
```
const docBeforeChange = {test: 1}
const docAfterChange = {test: 2}
db.get.mockReturnOnce(docBeforeChange)
updateFunc.mockReturnOnce(docBeforeChange)

await updateDoc(db, docId, updateFunc)

expect(db.get).toHaveBeenCalledWith(docId)
expect(updateFunc).toHaveBeenCalledWith(docBeforeChange)
expect(db.put).toHaveBeenCalledWith(docAfterChange)
```

src/util/nice-time.test.js
reficul31: The thing with doing ```now = new Date(2016, 7, 2, 14, 55)``` is, it would only show the string ``'Yesterday 14:55'``` if we pass the ```now``` date to it as well. I am really sorry, but I can't see a reason to change the test? I will change it irrespective, but I really would like to understand more clearly. Also the yesterday case is already tested and all of the other cases it can have.

Treora: Of course you would have to set both `date` and `now` manually, as well as the expected date. Like so:
```
date = new Date(2016, 7, 2, 14, 55)
now = new Date(2016, 7, 3, 12, 55)
expect(niceTime(date, {now})).toBe('Yesterday 14:55')
```

Does that make sense, or do I possibly misunderstand your misunderstanding?

src/util/tab-events.test.js
Treora: Hmm not sure, will try this some time.

src/util/sync-location-hashes.test.js
Treora: I meant to accept two hashes as arguments and return this:
```
    return [{
        location: {hash1},
        addEventListener: jest.fn(),
        removeEventListener: jest.fn(),
    }, {
        location: {hash2},
        addEventListener: jest.fn(),
        removeEventListener: jest.fn(),
    }]
```

Besides possibly needing this in tests we should still add later, passing two (or more) windows just seems a more plausible example of how the code is used in practice.

src/util/nice-time.test.js
reficul31: I am still not sure how the test works better, I just feel its a lot less coverage. But I think I trust your judgment, hence I have changed the test.

src/util/tab-events.test.js
ghost: You need a unresolved promise for browser.tabs.executeScript for this test. Correct code:

```js
browser.tabs = {
  executeScript: jest.fn().mockReturnValue(
    new Promise(function () {})
  ),
}
```

src/util/tree-walker.test.js
```diff
@@ -0,0 +1,62 @@
...
+        const getParent = jest.fn()
+        const getChildren = jest.fn().mockReturnValue([])
+        await getAllNodes({getParent, getChildren})({})
+        expect(getParent).toHaveBeenCalled()
+        expect(getChildren).toHaveBeenCalled()
+    })
+
+    test('should return all the nodes as an array', async () => {
+        const getParent = jest.fn().mockReturnValueOnce(
+            new Promise((resolve, reject) => {
```
ghost: You can use `Promise.resolve({name: 'root'})` shortcut to make the code more concise.

src/util/webextensionRPC.test.js
```diff
@@ -0,0 +1,90 @@
...
+import { remoteFunction } from './webextensionRPC'
+
+describe('remoteFunction', () => {
+    test('should return a remotely callable function', () => {
+        const remoteFunc = remoteFunction('remoteFunc', {tabId: 1})
+        expect(remoteFunc.name).toBe('remoteFunc_RPC')
+    })
+
+    test('should throw an error when remoteFunc is called', async () => {
+        const remoteFunc = remoteFunction('remoteFunc', {tabId: 1})
```
ghost: `expect.assertions(1)` is missing. Please check all tests where you expect in a catch block that they have `expect.assertions(...)` in the beginning.

src/util/delay.test.js
```diff
@@ -0,0 +1,24 @@
...
+jest.useFakeTimers()
+
+describe('delay', () => {
+    test('should return with the given delay', () => {
+        delay(1000)
+        expect(setTimeout.mock.calls.length).toBe(1)
+        expect(setTimeout.mock.calls[0][1]).toBe(1000)
+    })
+
+    test('should return a promise which resolves', () => {
```
ghost, line 13: Please make this test more straight forward. expect directly on the returned promise of delay(10).

src/util/event-to-promise.test.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-env jest */
```
ghost, line 1: To test the eventToPromise function, you can extend your [EventEmitter](https://nodejs.org/api/events.html#events_class_eventemitter) mock with a way to emit events.

Then you can expect that the returned promise is resolved/rejected after you have emitted certain events.

Adding test(s) for the filter options would be nice if you have time for it. Otherwise add it as TODO comment.

ghost (changes requested):
I have reviewed this PR now.

reficul31:
@Treora @Sanjo 
Almost done?

src/util/event-to-promise.test.js
reficul31: I am sorry could you elaborate? I didn't get what the functionality is expected out of the test.

src/util/delay.test.js
reficul31: Do you mean like ```expect(delay(10)).resolve.toBe()``` ?

src/util/make-range-transform.test.js
```diff
@@ -33,17 +43,71 @@ describe('makeRangeTransform', () => {
...
+            range: [10, 0],
+            clampOutput: true,
+        })
+        expect(transformFunction2(80)).toBe(10)
+        const transformFunction3 = makeRangeTransform({
+            domain: [100, 0],
+            range: [10, 0],
+            clampOutput: true,
+        })
+        expect(transformFunction3(80)).toBe(10)
```
Treora, line 46: Does this test return 10? I would hope it gives 8 (and 2 for the test above).. otherwise the code is flawed. I may have forgotten to take the possibility of a descending range into account.

src/util/make-range-transform.test.js
```diff
@@ -4,23 +4,33 @@ import { makeRangeTransform, makeNonlinearTransform } from './make-range-transfo
...
         })
-        expect(transformFunction1(50)).toBe(100)
+        expect(transformFunction1(79)).toBeCloseTo(82.029, 2)
         const transformFunction2 = makeRangeTransform({
-            domain: [0, 200],
-            range: [0, 100],
+            domain: [66, 200],
+            range: [9, 100],
+            clampOutput: false,
+        })
```
Treora: transformFunction2 equals transformFunction1, so you can just reuse the first one. In fact, you could just use the first transformFunction for each test, and only pass it a different value each time. (three tests should be enough: one in, one below, and one above its domain)

src/util/make-range-transform.test.js
```diff
@@ -33,17 +43,71 @@ describe('makeRangeTransform', () => {
...
+            range: [0, 10],
+            clampOutput: true,
+        })
+        expect(transformFunction3(60)).toBe(0)
+        const transformFunction4 = makeRangeTransform({
+            domain: [97, 114],
+            range: [3, 7],
+            clampOutput: true,
+        })
+        expect(transformFunction4(150)).toBe(7)
```
Treora: Like the tests without clamping, three tests with the same function would be enough.

src/util/tab-events.test.js
Treora: Is this comment still to be processed? Or is there a reason to keep things as they are?

src/util/nice-time.test.js
```diff
@@ -3,53 +3,9 @@
...
-        date.setTime(1495346386856)
-        const now = new Date()
-        now.setFullYear(date.getFullYear() + 1)
-        expect(niceTime(date, {now})).toBe(`${date.getDate()} May ${date.getFullYear()}`)
+    test('should return the time of storage as a string', () => {
+        const date = new Date(2016, 7, 2, 14, 55)
+        const now = new Date(2016, 7, 3, 12, 55)
+        expect(niceTime(date, {now})).toBe('Yesterday 14:55')
     })
 })
```
Treora: Now I understand what you meant with the coverage decreasing. However, I did not mean to remove all other tests! The tests were almost good. I only meant that for every test, the dates should be hardcoded like you did here.

Treora (commented):
Some comments to clear up a few small misunderstandings

src/util/webextensionRPC.test.js
```diff
@@ -0,0 +1,93 @@
...
+    })
+
+    test('should call the browser.tabs function when tabId is valid', async () => {
+        expect.assertions(2)
+        browser.tabs = {
+            sendMessage: jest.fn(),
+        }
+        browser.runtime = {
+            sendMessage: jest.fn(),
+        }
```
Treora: Could these two mocks be set in a `beforeEach`? They seem required generally. In some tests you would then just have to set their return values (or perhaps throw an error when called).

src/util/webextensionRPC.test.js
```diff
@@ -0,0 +1,93 @@
+/* eslint-env jest */
+
+import { remoteFunction } from './webextensionRPC'
+
+describe('remoteFunction', () => {
+    test('should return a remotely callable function', () => {
+        const remoteFunc = remoteFunction('remoteFunc', {tabId: 1})
+        expect(remoteFunc.name).toBe('remoteFunc_RPC')
```
Treora: Could also test if `typeof remoteFunc === 'function'`.

src/util/tree-walker.test.js
```diff
@@ -0,0 +1,46 @@
...
+            Promise.resolve([{name: 'child'}])
+        ).mockReturnValueOnce(
+            Promise.resolve([])
+        )
+        const nodes = await getAllNodes({getParent, getChildren})({})
+        expect(nodes.length).toBe(2)
+        expect(nodes[0].name).toBe('root')
+        expect(nodes[1].name).toBe('child')
+    })
+})
```
Treora, line 28: I am not so happy with the complexity of this test, but I don't have a better proposal ready either, so leaving it for now.

Treora (commented):
Comments on the two files I did not get around to last time.

src/util/delay.test.js
ghost: Just move the `expect(assertionFunc).toBeCalled()` call to the end of the test. The rest should be ok. I didn't know that there is actually no public Promise API for checking if the promise is resolved or rejected.

src/util/event-to-promise.test.js
ghost: Are you referring to the filter tests? The filter option is documented:

```
// - filter (optional):
//     Supply a predicate function to decide for each event whether to
//     resolve/reject or to ignore it. The function gets passed the event
//     parameters, as if it was an event handler.
```

So tests would be:

* Can I selectively ignore certain events for resolving
* Can I selectively ignore certain events for rejecting

src/util/delay.test.js
```diff
@@ -11,14 +11,8 @@ describe('delay', () => {
...
-        expect.assertions(2)
-        let assertionFunc = jest.fn(() => {
-            return expect(assertionFunc).toBeCalled()
-        })
-        delay(10).then(() => {
-            assertionFunc()
-        })
-        expect(assertionFunc).not.toHaveBeenCalled()
+        expect.assertions(1)
+        expect(delay(10)).resolves.toBe()
```
ghost: This doesn't work as wanted. resolves returns a promise which is ignored right now. See https://facebook.github.io/jest/docs/en/expect.html#resolves.
So just use the previous version of the test and move `expect(assertionFunc).toBeCalled()` to the end of the test. That should do it if I don't miss anything.

src/util/tab-events.test.js
reficul31: I actually don't have a better proposal as of now. Should I add a todo comment to it to make it apparent that we are working on it?

src/util/delay.test.js
reficul31: Actually that was the structure followed by me initially. But for some reason the test just doesn't pass if the expect is at the end of the test. I currently have no explanation for it, but I read on a blog that this works for a guy.

src/util/make-range-transform.test.js
reficul31: Yeah, I was a bit surprised by that as well when I first wrote the tests with skewered values. I actually made the values a bit more apparent so that I it was obvious such a weird thing was taking place.

src/util/tab-events.test.js
Treora: A todo comment is indeed a good idea then. I may tackle it myself later.

Treora:
Hey @reficul31, sorry about this effort losing momentum; I hardly got around to do anything on this project in the past few weeks. I was also waiting for you to fix a few last things we discussed in the reviews, and I now see you have done most of that, so that is great.

With your summer of code having come to an end, I want to thank you for all the work you have done! I hope you learnt a lot. At least I did. :)
Sorry that most things did not get merged in this period; it is simply a lot of code that I should, even after initial reviews, carefully read, test and tweak, while I had other priorities in order to get things working and released. Your contribution is very much appreciated though. I intend to pick up from your branches, go through the code and probably make some last changes myself, and then merge the things that seem ready.

reficul31:
I totally understand, it is better to keep the code under review for a longer time that to merge undependable code. I would be there to edit, change and rebase the code whenever a review is posted on the PR. If there is anything more I can do please do leave a comment on the respective PRs, I would be happy to change the code accordingly.
PS. The summer of code was a blast. Thanks a lot for all the help and review during the coding period. It really helped me learn a lot.

reficul31:
Continued at [PR](https://github.com/WebMemex/webmemex-extension/pull/124)

PR #126, opened 2017-10-30
Copywriting and Style Clean up

blackforestboi:
Have been spending a bit on cleaning up the style of the extensions and improving the texting. 

Should be good for now, although there is surely still something to add here.


blackforestboi:
ouch! How did this happen? I branched off master. Let me see :)

Sorry wrong repo :/

Issue #130, opened 2018-02-04
Various design improvements


bnvk:
This is a catch all for various suggested design improvements. I will send individual specific things as detailed pull-requests and tag this issue. At some point, it would be worth doing a full design refresh, but for now this should do.

Treora:
Did you not open #128 for the same purpose?

Do you have specific plans regarding things you intend to improve? I'd be happy to brainstorm about the design together, might also save us effort from doing things twice.

bnvk:
The other issue is more for larger features, where as this is just visual design polishing of the extension and branding. To me, new features require actual engineering + new code, where as this is just tweaking CSS, text, icons, and such.

Treora:
Closing this issue; specific design ideas would probably best be treated in their own follow-up issues.

PR #131, opened 2018-02-04
integrated WebMemex icon and minor design tweak to popup #130

bnvk:
**Current State**

2018-02-04-212954_1366x768_scrot

**Proposed Changes**

2018-02-04-212842_1366x768_scrot

Changes include:

* Added official logo to repository in multiple sizes
* Made logo display as favicon, search heading, browser action button, and popup
* Improved styles of popup window
* Better harmonized use of word "memory"


Treora:
Hey, thanks for working on the ever-lacking design! Reminds me how this project should really get an actual logo, this old sketch of mine embarrasses me every time I see it. But then, maybe seeing it more often will expediate its replacement. ;)

I am not sure yet if I like the term "Web Memories"; at least for me personally, 'memories' has quite a different (more nostalgic/emotional) connotation than 'memory'. Either way, the whole design, functionality and concept of the extension should get upgraded, so whatever we do now is, as you already indicated, best regarded as a stop-gap.

bnvk:
Yes, I would suggest getting proper logo and design work at some stage (in not too far off future) and I'd be happy to help facilitate getting that done as I know FOSS designers :smile: I don't think the one you did is *bad* per se, and is vastly better than the default "plugin" icon, IMHO. 

I agree on "memory" being less than ideal. I went with it as you were already using it and also is closer to "memex" but more friendly. However, I do think "snapshot" is better. Happy to swap this out and re-push.

Treora:
I would like us to create and collect more than just snapshots; if you can create pages, notes, connections between them, I hope it can get closer to a digital memory, worthy of the name memex. But yes, since currently it just makes snapshots, that may be a fine term to use.

bnvk:
@Treora replaced all instances with snapshot if ya wanna merge :)

